PR #4, opened 2025-03-02
Implement `ret`

Smaug123:
After this PR, the implementation of HelloWorld incorrectly defers loading the Program module until after the `main` call (that's a behavioural bug, although not one that should affect most users), but then we do the following:

* Call `reallyMain`
* Identify that we are loading `System.Console.WriteLine`, looking it up and discovering that it is forwarded to System.Private.CoreLib
* Identify the base class of `System.Console` (namely `obj`), call its static constructor (there is none), call the static constructor of `System.Console` which has three opcodes (`newobj`, `stsfld`, and `ret`: this allocates and stores [a single reference type](https://github.com/dotnet/runtime/blob/5535e31a712343a63f5d7d796cd874e563e5ac14/src/libraries/System.Console/src/System/Console.cs#L27), the others being value types which should be initialised to their default values, but I haven't implemented that bit yet)
* Imperfectly call `newobj`, which in this case is constructing a `System.Object`; that is simply a `ret` opcode, which we execute. Note that we are supposed to have performed an allocation, and I haven't done that yet!
* Identify that we're calling `stsfld`, and throw-with-TODO.

### DIFF
--- a/WoofWare.PawPrint/IlOp.fs
+++ b/WoofWare.PawPrint/IlOp.fs
@@ -239,6 +239,10 @@ type NullaryIlOp =
     | Readonly
     | Refanytype
 
+    /// The number of bytes this instruction takes in memory.
+    static member NumberOfBytes (op : NullaryIlOp) : int =
+        TODO
+
 type UnaryConstIlOp =
     | Stloc of uint16
     | Stloc_s of int8
@@ -287,6 +291,11 @@ type UnaryConstIlOp =
     | Ldloca of uint16
     | Ldarg of uint16
 
+    /// The number of bytes this instruction takes in memory, including its constant argument that is inline in the
+    /// byte stream.
+    static member NumberOfBytes (op : UnaryConstIlOp) : int =
+        TODO: check the number of bytes in the argument, and check the special list of multi-byte ops below, and add the count.
+
 type UnaryMetadataTokenIlOp =
     | Call
     | Calli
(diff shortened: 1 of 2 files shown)


WoofWare.PawPrint/IlOp.fs
```diff
@@ -331,37 +350,34 @@ type UnaryMetadataTokenIlOp =
...
-0xFE 0x17 	cpblk 	Copy data from memory to memory. 	Base instruction
-0xFE 0x18 	initblk 	Set all bytes in a block of memory to a given byte value. 	Base instruction
-0xFE 0x19 	no 	The specified fault check(s) normally performed as part of the execution of the subsequent instruction can/shall be skipped. 	Prefix to instruction
-0xFE 0x1A 	rethrow 	Rethrow the current exception. 	Object model instruction
-0xFE 0x1C 	sizeof <typeTok> 	Push the size, in bytes, of a type as an unsigned int32. 	Object model instruction
-0xFE 0x1D 	refanytype 	Push the type token stored in a typed reference. 	Object model instruction
-0xFE 0x1E 	readonly. 	Specify that the subsequent array address operation performs no type check at runtime, and that it returns a controlled-mutability managed pointer. 	Prefix to instruction
+        match op with
+        | Ldftn | Ldvirtftn | Initobj | Constrained | Sizeof -> 2 + 4 // Two-byte opcode + 4-byte token
+        | Call
```
Smaug123, line 355: Claude gave me an underscore here, but I used Rider's quick action and then a vim macro to replace the underscore with all the types, just in case my list is currently incomplete and I end up adding to it.

PR #6, opened 2025-05-16
Expand test harness

Smaug123:
Uses WoofWare.DotnetRuntimeLocator to find the appropriate runtime.


.github/workflows/dotnet.yaml
```diff
@@ -31,8 +31,10 @@ jobs:
       run: nix develop --command dotnet restore
     - name: Build
       run: nix develop --command dotnet build --no-restore --configuration Release
+    - name: Publish
+      run: nix develop --command dotnet publish --configuration Release --runtime linux-x64 --self-contained
```
Smaug123, line 35: Not actually necessary, just makes it a bit simpler.

PR #10, opened 2025-05-19
Add extremely basic support for generics

Smaug123:


CSharpExample/Class1.cs
```diff
@@ -1,17 +1,25 @@
 ﻿using System;
 using System.IO;
+using System.Threading;
```
Smaug123, line 3: This just gets rid of the syntactic sugar and also stops including `new FileInfo`.

PR #15, opened 2025-05-26
Add hook for non-IL methods

Smaug123:


Smaug123:
I appear to be unable to turn off Copilot after enabling it yesterday in a spirit of investigation.

PR #38, opened 2025-06-02
Better generics handling

Smaug123:
I say "better", but what I really mean is "more complicated".

### DIFF
--- a/WoofWare.PawPrint/IlMachineState.fs
+++ b/WoofWare.PawPrint/IlMachineState.fs
@@ -276,6 +276,7 @@ module IlMachineState =
 
     let rec resolveTypeFromDefn
         (loggerFactory : ILoggerFactory)
+        (corelib : BaseClassTypes<DumpedAssembly>)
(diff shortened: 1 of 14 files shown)


WoofWare.PawPrint/IlMachineState.fs
```diff
@@ -298,17 +336,48 @@ module IlMachineState =
...
+                | PrimitiveType.Object -> failwith "todo"
+                |> TypeInfo.mapGeneric (fun _ -> failwith "none of these types are generic")
+
+            state, corelib.Corelib, ty
+        | TypeDefn.GenericTypeParameter param ->
+            match genericArgs with
+            | None -> failwith "tried to resolve generic parameter without generic args"
+            | Some genericArgs ->
+                let arg = genericArgs.[param]
+                // TODO: this assembly is probably wrong?
```
Smaug123, line 367: Yeah, this indicates that we should simply not be handing assemblies round all over the place. They should be contained entirely within the TypeDefn.

PR #49, opened 2025-06-15
Ldtoken

Smaug123:


WoofWare.PawPrint/UnaryMetadataIlOp.fs
```diff
@@ -940,7 +939,53 @@ module internal UnaryMetadataIlOp =
...
         | Stobj -> failwith "TODO: Stobj unimplemented"
         | Constrained -> failwith "TODO: Constrained unimplemented"
-        | Ldtoken -> failwith "TODO: Ldtoken unimplemented"
+        | Ldtoken ->
+            let state =
+                match metadataToken with
+                | MetadataToken.FieldDefinition h ->
+                    let ty = baseClassTypes.RuntimeFieldHandle
+                    // this is a struct; it contains one field, an IRuntimeFieldInfo
+                    // https://github.com/dotnet/runtime/blob/1d1bf92fcf43aa6981804dc53c5174445069c9e4/src/coreclr/System.Private.CoreLib/src/System/RuntimeHandles.cs#L1097
```
Smaug123, line 948: The type may be known to the C++ runtime, but I don't yet know. https://github.com/dotnet/runtime/blob/f0168ee80ba9aca18a7e7140b2bb436defda623c/src/coreclr/System.Private.CoreLib/src/System/RuntimeHandles.cs#L1967 ; hopefully tracking the implementation down we might discover that it creates https://github.com/dotnet/runtime/blob/f0168ee80ba9aca18a7e7140b2bb436defda623c/src/coreclr/System.Private.CoreLib/src/System/Reflection/RtFieldInfo.cs#L12

PR #99, opened 2025-08-11
Implement Unsafe.As

Smaug123:


Smaug123:
Just going to fold this into #96.

PR #136, opened 2026-04-09
Implement phase 0-4 of type identity refactor

Smaug123:
GPT-5.4 xhigh and Opus 4.6, with GPT-5.2 xhigh also involved in reviews.


WoofWare.PawPrint.Test/TestTypeResolution.fs
```diff
@@ -0,0 +1,729 @@
...
+    type private NoAssemblyLoad () =
+        interface IAssemblyLoad with
+            member _.LoadAssembly _loadedAssemblies _referencedIn _handle =
+                failwith "Test unexpectedly attempted to load an assembly"
+
+    let private loggerFactory () = LoggerFactory.makeTest () |> snd
+
+    let private dumpedAssembly (path : string option) (bytes : byte[]) : DumpedAssembly =
+        use stream = new MemoryStream (bytes)
+        global.WoofWare.PawPrint.AssemblyApi.read (loggerFactory ()) path stream
```
Smaug123, line 54: Good lord I never knew this syntax

Smaug123:
Review from Robocop rerun synchronously:

```
* **Bug: concretisation load/retry logic breaks nested TypeRefs + multi-step loads** — `WoofWare.PawPrint.Domain/TypeConcretisation.fs`, `loadAssemblyAndResolveTypeRef` ignores the `assemblyRef` from `TypeResolutionResult.FirstLoadAssy` and instead matches on `typeRef.ResolutionScope` (only handles `TypeRefResolutionScope.Assembly`, else `failwith "Unexpected resolution scope"`), and also fails if a second `FirstLoadAssy` is needed.
  * Context: `| TypeResolutionResult.FirstLoadAssy assemblyRef -> match typeRef.ResolutionScope with | TypeRefResolutionScope.Assembly assyRef -> … | _ -> failwith "Unexpected resolution scope"`.
* **Likely regression: chained type forwarders won’t resolve** — `WoofWare.PawPrint.Domain/Assembly.fs`, `resolveTypeFromExport` (`ExportedTypeData.ForwardsTo`) uses `resolveExportedTypeByChain` which only searches `TryGetTopLevelTypeDef`/`TryGetNestedTypeDef` in the forwarded-to assembly. If that assembly forwards/exports again, resolution now fails instead of continuing the chain.
  * Context: `| ExportedTypeData.ForwardsTo assyRef -> … | true, toAssy -> let identity = resolveExportedTypeByChain toAssy None ty`.
* **Interpreter retry path for exported types drops nested-export context** — `WoofWare.PawPrint/IlMachineState.fs`, `resolveTypeFromExport` loads the requested assembly and then calls `resolveTypeFromName loggerFactory ty.Namespace ty.Name …`. This won’t work for nested exported types (`ty.Namespace = None`) and bypasses the new exported-parent-chain resolver; it should retry by calling `resolveTypeFromExport` again after loading.
  * Context: `| TypeResolutionResult.FirstLoadAssy loadFirst -> … resolveTypeFromName loggerFactory ty.Namespace ty.Name …`.
```

Smaug123:
```
  - [P1] Fall back to nested exported types when child typedef is absent — /Users/patrick/Documents/GitHub/
    WoofWare.PawPrint2/WoofWare.PawPrint.Domain/Assembly.fs:559-560
    For nested type refs that traverse a forwarding assembly, this helper now only probes
    TryGetNestedTypeDef. If Outer resolves through a top-level export and Inner is represented by an
    ExportedTypeData.ParentExportedType row instead of a local nested TypeDef, resolution still throws here
    even though the patch added _NestedExportedTypesLookup and resolveTypeFromExport support for that
    metadata. In practice, Outer.Inner references through nested type forwarders remain broken because
    production code never consults the nested exported-type lookup.
  - [P3] Invalidate ambiguous top-level exported-type keys again — /Users/patrick/Documents/GitHub/
    WoofWare.PawPrint2/WoofWare.PawPrint.Domain/Assembly.fs:222-230
    If an assembly contains more than one top-level ExportedType row for the same (namespace, name) (for
    example, duplicate forwarders or a forwarder plus another export), this builder now leaves the first
    row in the lookup and only logs the duplicate. The previous implementation removed ambiguous keys, so
    callers failed explicitly instead of resolving whichever row happened to appear first in metadata
    order. With the new code, forwarded-type resolution can silently pick the wrong target assembly/module
    in those inputs.
```

I'd like to block further work on this branch, though, pending a new comprehensive PBT-based approach to testing.

PR #142, opened 2026-04-11
Stop keying static fields by name

Smaug123:
Not a correctness issue, but a bit more safe for the future.

Opus 4.6.


WoofWare.PawPrint.Test/sourcesPure/StaticFieldShadowing.cs
```diff
@@ -0,0 +1,31 @@
+class Base
```
Smaug123, line 1: This test currently passes on main anyway.

PR #161, opened 2026-04-16
Migrate containsRefType walker to TypeInfo.isReferenceType

Smaug123:
Stage 3c of the type-predicate refactor in
docs/plans/2026-04-16-type-predicates.md. The Intrinsics.containsRefType short-circuit asks "is this type itself a reference type?" — replace the hand-rolled `match resolveBaseType` with the intent-named predicate. Pure refactor.


Smaug123:

PR #163, opened 2026-04-16
Migrate delegate-dispatch branch to TypeInfo.isDelegate

Smaug123:
Stage 3d of the type-predicate refactor in
docs/plans/2026-04-16-type-predicates.md. AbstractMachine's no-IL path dispatches to delegate handling iff the executing method's declaring type is a concrete delegate. `isDelegate` encodes exactly that (excluding exact System.Delegate and System.MulticastDelegate, neither of which should take the user-delegate ctor/invoke branch). Pure refactor.


Smaug123:

PR #164, opened 2026-04-16
Migrate IsReferenceOrContainsReferences to TypeInfo.isValueType

Smaug123:
Stage 3e of the type-predicate refactor in
docs/plans/2026-04-16-type-predicates.md. The intrinsic for RuntimeHelpers.IsReferenceOrContainsReferences<T> asks "is T a value type (recurse into fields) or a reference type (answer true)?" — the intent-named predicate. Minor behaviour change: if T is exact System.Enum, the answer flips from "recurse into fields" to "return true", which matches real .NET (typeof(Enum).IsValueType is false).


Smaug123:

PR #170, opened 2026-04-17
Select the innermost exception handler

Smaug123:


WoofWare.PawPrint/ExceptionDispatching.fs
```diff
@@ -107,10 +107,34 @@ module ExceptionDispatching =
                         state, acc
             )
 
+        // When multiple regions match (e.g. a catch and a finally for the same try block),
+        // pick the innermost (smallest TryLength) handler. Among equal-sized try regions,
+        // prefer catch over finally/fault per ECMA-335 §I.12.4.2.
```
Smaug123: Specifically I.12.4.2.7, although it's not stated quite that way.

PR #180, opened 2026-04-18
Implement box IL instruction for value types and Enum.HasFlag intrinsic

Smaug123:
Boxing value types was previously unimplemented, causing any code that boxes enums or structs to fail. The box instruction now allocates a heap object containing the value type's fields and pushes an ObjectRef.

Also adds Enum.HasFlag as a JIT intrinsic (https://github.com/dotnet/runtime/blob/dbd3e33df9ccf74b91045e095477726c2bf83916/src/libraries/System.Private.CoreLib/src/System/Enum.cs#L398), extracting the underlying integer from each boxed enum and performing the bitwise AND check.


WoofWare.PawPrint/Intrinsics.fs
```diff
@@ -732,5 +732,42 @@ module Intrinsics =
             |> IlMachineState.pushToEvalStack' toPush currentThread
             |> IlMachineState.advanceProgramCounter currentThread
             |> Some
+        | "System.Private.CoreLib", "Enum", "HasFlag" ->
```
Smaug123, line 735:
```suggestion
        | "System.Private.CoreLib", "Enum", "HasFlag" ->
            // https://github.com/dotnet/runtime/blob/dbd3e33df9ccf74b91045e095477726c2bf83916/src/libraries/System.Private.CoreLib/src/System/Enum.cs#L398
```

PR #200, opened 2026-04-20
Implement Unsafe.{Add,AreSame,ByteOffset} array-byref intrinsics

Smaug123:
(Claude Opus 4.7.)

These three JIT intrinsics are the pointer-arithmetic prerequisites for running the BCL's Buffer.Memmove / SpanHelpers.Memmove IL faithfully (see docs/plans/2026-04-20-memmove.md, PR A). Each intrinsic's managed body throws PlatformNotSupportedException at runtime; the JIT normally rewrites them. Here we handle them on the PawPrint side:

- Unsafe.AreSame(ref T, ref T): structural equality on ManagedPointerSource.
- Unsafe.Add(ref T, int): shift an ArrayElement byref by the given offset.
- Unsafe.ByteOffset(ref T, ref T): (i2 - i1) * sizeof(T) between two byrefs into the same array, boxed into an IntPtr ValueType.

All three restrict to Byref(ArrayElement arr i, []) shapes for now and fail loudly on anything richer (projections, cross-array ByteOffset, non-int32 Add overloads). That's the minimum needed for the Memmove IL path and keeps the surface easy to extend when richer byrefs show up.

Also extends ceq's NativeInt/NativeInt arm to treat Verbatim 0L as equal to ManagedPointer Null, since arithmetic zero and the IntPtr.Zero literal are both valid representations of zero but used different NativeIntSource variants.

### DIFF
--- a/WoofWare.PawPrint/Intrinsics.fs
+++ b/WoofWare.PawPrint/Intrinsics.fs
@@ -664,6 +664,11 @@ module Intrinsics =
                 | EvalStackValue.Float _ -> failwith "expected pointer type"
                 | EvalStackValue.NativeInt nativeIntSource -> failwith "todo"
                 | EvalStackValue.NullObjectRef -> failwith "todo: Unsafe.As on null"
+                | EvalStackValue.ManagedPointer src when from = to_ ->
+                    // Unsafe.As<T,T> is a no-op: same address and same type view.
+                    // Skipping the projection keeps the representation canonical so
+                    // that AreSame / ceq on the result compares equal to the input.
+                    EvalStackValue.ManagedPointer src
                 | EvalStackValue.ManagedPointer src ->
                     ManagedPointerSource.appendProjection (ByrefProjection.ReinterpretAs to_) src
                     |> EvalStackValue.ManagedPointer
--- a/WoofWare.PawPrint/Intrinsics.fs
+++ b/WoofWare.PawPrint/Intrinsics.fs
@@ -750,8 +750,18 @@ module Intrinsics =
                 | EvalStackValue.Int32 i -> i
                 | _ -> failwith $"TODO: Unsafe.Add: expected Int32 offset, got %O{offset}"
 
-            let ptr : EvalStackValue =
+            // Strip trailing address-preserving `ReinterpretAs` projections: a
+            // round-trip `Unsafe.As<U,T>(Unsafe.As<T,U>(x))` leaves a collapsed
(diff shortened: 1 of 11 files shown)


WoofWare.PawPrint/Intrinsics.fs
```diff
@@ -695,6 +695,107 @@ module Intrinsics =
...
+                | _ -> failwith $"TODO: Unsafe.AreSame: expected ManagedPointer, got %O{v}"
+
+            let areSame = (extractPtr left) = (extractPtr right)
+
+            state
+            |> IlMachineState.pushToEvalStack (CliType.ofBool areSame) currentThread
+            |> IlMachineState.advanceProgramCounter currentThread
+            |> Some
+        | "System.Private.CoreLib", "Unsafe", "Add" ->
+            // https://github.com/dotnet/runtime/blob/108fa7856efcfd39bc991c2d849eabbf7ba5989c/src/coreclr/tools/Common/TypeSystem/IL/Stubs/UnsafeIntrinsics.cs#L192
```
Smaug123:
```suggestion
            // https://github.com/dotnet/runtime/blob/108fa7856efcfd39bc991c2d849eabbf7ba5989c/src/coreclr/tools/Common/TypeSystem/IL/Stubs/UnsafeIntrinsics.cs#L99
```

WoofWare.PawPrint/Intrinsics.fs
```diff
@@ -695,6 +695,107 @@ module Intrinsics =
...
+                    ManagedPointerSource.Byref (ByrefRoot.ArrayElement (arr, i + offset), [])
+                    |> EvalStackValue.ManagedPointer
+                | _ -> failwith $"TODO: Unsafe.Add on non-plain-array-element byref: %O{src}"
+
+            state
+            |> IlMachineState.pushToEvalStack' ptr currentThread
+            |> IlMachineState.advanceProgramCounter currentThread
+            |> Some
+        | "System.Private.CoreLib", "Unsafe", "ByteOffset" ->
+            // https://github.com/dotnet/runtime/blob/108fa7856efcfd39bc991c2d849eabbf7ba5989c/src/coreclr/tools/Common/TypeSystem/IL/Stubs/UnsafeIntrinsics.cs#L142
```
Smaug123:
```suggestion
            // https://github.com/dotnet/runtime/blob/108fa7856efcfd39bc991c2d849eabbf7ba5989c/src/coreclr/tools/Common/TypeSystem/IL/Stubs/UnsafeIntrinsics.cs#L69
```

PR #205, opened 2026-04-20
Populate TypeInitializationException fields on .cctor failure

Smaug123:
When a .cctor throws, synthesise the wrapping TIE with its _typeName, _HResult, and _innerException set. This brings the observable shape of the exception closer into line with what the CLR produces, though we still don't give you the StackTrace property.

The ThrowingCctorProperties and ThrowingCctorStackTrace tests are added to unimplementedMockTests: even after setting StackTrace (which we don't do here), the TIE fields are correctly populated, but end-to-end execution hits RuntimeTypeHandle::GetGCHandle (PInvoke) in a deeper BCL path that is not yet implemented.

### DIFF
--- a/WoofWare.PawPrint/ExceptionDispatching.fs
+++ b/WoofWare.PawPrint/ExceptionDispatching.fs
@@ -14,57 +14,6 @@ type ExceptionDispatchResult =
 [<RequireQualifiedAccess>]
 module ExceptionDispatching =
 
-    /// Format a single stack frame in a CLR-like style:
-    ///   "   at <namespace>.<type>.<method>()"
-    /// We don't render argument lists or file info, but the frames carry enough identifying
-    /// detail for managed code to locate the caller (.cctor, Main, etc.) via string search.
-    let private formatStackFrame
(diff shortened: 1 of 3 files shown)


WoofWare.PawPrint/ExceptionDispatching.fs
```diff
@@ -14,6 +14,57 @@ type ExceptionDispatchResult =
...
+                sprintf "%s.%s" declaringType.Namespace declaringType.Name
+
+        sprintf "   at %s.%s()" typeFullName frame.Method.Name
+
+    /// Format a list of stack frames, innermost first, one per line (no trailing newline).
+    let private formatStackTrace
+        (frames : ExceptionStackFrame<ConcreteTypeHandle, ConcreteTypeHandle, ConcreteTypeHandle> list)
+        : string
+        =
+        frames |> List.map formatStackFrame |> String.concat System.Environment.NewLine
```
Smaug123: Claude notes that this is a guest-observable property of the host. I'm fine with that for now, but raised #206.

PR #211, opened 2026-04-21
Add nominal primitive-like struct registry

Smaug123:
We've got a couple of functions that interpret "struct with one field" interchangeably with "the field", and this is really nasty. Add a registry so we can distinguish them when the CLR treats them interchangeably.

### DIFF
--- a/WoofWare.PawPrint.Domain/TypeInfo.fs
+++ b/WoofWare.PawPrint.Domain/TypeInfo.fs
@@ -192,6 +192,9 @@ type BaseClassTypes<'corelib> =
         TypedReference : TypeInfo<GenericParamFromMetadata, TypeDefn>
         IntPtr : TypeInfo<GenericParamFromMetadata, TypeDefn>
         UIntPtr : TypeInfo<GenericParamFromMetadata, TypeDefn>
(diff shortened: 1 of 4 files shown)


WoofWare.PawPrint.Domain/TypeInfo.fs
```diff
@@ -192,6 +192,9 @@ type BaseClassTypes<'corelib> =
         TypedReference : TypeInfo<GenericParamFromMetadata, TypeDefn>
         IntPtr : TypeInfo<GenericParamFromMetadata, TypeDefn>
         UIntPtr : TypeInfo<GenericParamFromMetadata, TypeDefn>
+        /// `System.ByReference` (non-generic in modern corelibs) or `System.ByReference<T>` (older).
+        /// Optional because not every supported corelib exposes it.
+        ByReference : TypeInfo<GenericParamFromMetadata, TypeDefn> option
```
Smaug123, line 197: https://github.com/dotnet/runtime/blob/81b1c762c37d15a57a3b3b693447fbcd8d6df55c/src/libraries/System.Private.CoreLib/src/System/ByReference.cs#L12

PR #230, opened 2026-04-21
Flatten enums on the eval stack

Smaug123:


WoofWare.PawPrint/BasicCliType.fs
```diff
@@ -506,6 +506,14 @@ and CliValueType =
             /// the context-free `EvalStackValue.ofCliType` can flatten without threading
             /// `BaseClassTypes`/`AllConcreteTypes` through every push site.
             _PrimitiveLikeKind : PrimitiveLikeKind option
+            /// Cached enum classification: `true` iff this value type is the storage shape of a CLR
+            /// enum (exactly one instance field at offset 0 named `value__` with an integral
+            /// underlying type). Unlike primitive-likes, enums stay as `UserDefinedValueType` on
+            /// the eval stack; they only need a nominal handle on the pop-side rewrap and on `ceq`.
+            /// The `value__` name is CLR-reserved for enums, so this structural test is equivalent
```
Smaug123, line 513: This is foul but Claude Opus 4.7's suggestions for fixing it are even worse.

```
  value__ is CLR-reserved (ECMA-335 II.22.15 — the enum's single instance field must be named value__, and
  user code can't declare a field with that name), so structurally it's a reliable marker. But "reliable"
  isn't the same as "right" — the nominal fact we actually want is "this type's base is System.Enum", and
  we're pattern-matching on a shape proxy instead of asking the type system.

  The reason it went in that way: ComputeIsEnumLike runs inside
  CliValueType.OfFields/AddField/WithFieldSet, which are pure record constructors. They don't currently
  take the BaseClassTypes / corelib handle needed to resolve "is this type's base System.Enum?" — so going
  nominal means threading that through every construction site, or deferring the classification until first
   use.

  Two cleaner options if you want to de-hack it:
  1. Nominal at construction: thread the corelib/base-type resolver into the constructors. Biggest blast
  radius, but the classification becomes a real nominal check.
  2. Classify at the seam, not in the type: drop _IsEnumLike from CliValueType entirely and make the
  eval-stack boundary ask a helper isEnumLike : TypeInfo -> bool that walks base types. The cache moves to
  where the lookup context already exists (the interpreter has corelib in hand at push/pop time).
```

PR #234, opened 2026-04-22
Refactor strings as char arrays in the array map

Smaug123:
Opus 4.7[1m] ultrareview:

🟡 WoofWare.PawPrint/UnaryMetadataIlOp.fs:1141
> The isStringFirstChar interception is missing from the ManagedPointer branches in the Ldflda, Ldfld, and Stfld handlers. If a ManagedPointer to a String somehow reached these paths, the code would attempt to project through _firstChar as a regular field, which no longer exists on the String heap object after this refactor. In practice this is unreachable because String is a reference type (so this is always an ObjectRef), but adding the guard to the ManagedPointer branches would make the interception consistent and robust against future changes.

### DIFF
--- a/WoofWare.PawPrint/IlMachineState.fs
+++ b/WoofWare.PawPrint/IlMachineState.fs
@@ -1923,32 +1923,6 @@ module IlMachineState =
         then
             failwith $"unexpectedly don't know how to initialise a string: got fields %O{stringInstanceFields}"
 
-        // Sibling char[] holds the full null-terminated char payload. We allocate it via the
-        // ordinary array allocator: the sibling is an entirely normal entry in `Arrays`, and
-        // its heap address goes into the `StringCharArrays` side-table so consumers can look
-        // it up from the String's address.
-        let charPayload =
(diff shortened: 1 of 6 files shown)


WoofWare.PawPrint/IlMachineState.fs
```diff
@@ -1947,12 +1961,6 @@ module IlMachineState =
...
         let fields =
             [
-                {
-                    Name = "_firstChar"
-                    Contents = CliType.ofChar state.ManagedHeap.StringArrayData.[dataAddr]
-                    Offset = None
-                    Type = AllConcreteTypes.getRequiredNonGenericHandle state.ConcreteTypes baseClassTypes.Char
-                }
                 {
                     Name = "_stringLength"
```
Smaug123, line 1929: This really is *mighty* spooky.

Smaug123 (commented):
I find this all exceptionally weird - why on earth are strings special-cased in this way. Surely there are other sibling-backed types.

PR #238, opened 2026-04-24
Resolve virtual overrides for constrained.callvirt

Smaug123:


Smaug123:
OK this is just totally wrong - we shouldn't be mocking out managed functions

PR #251, opened 2026-04-24
Enable the ConstrainedCallvirtArrayT test

Smaug123:


Smaug123:
This whole shape is wrong; special-casing arrays like this is bad.

PR #256, opened 2026-04-24
Concretize the main types before program load

Smaug123:


WoofWare.PawPrint/Program.fs
```diff
@@ -294,33 +294,36 @@ module Program =
...
+                ImmutableArray.Empty // No type generics for main method's declaring type
+                rawMainMethod
+                None
+                dumped.Name
+                ImmutableArray.Empty
+                state
+
+        let state =
+            { state with
+                ConcreteTypes = Corelib.concretizeAll state._LoadedAssemblies baseClassTypes state.ConcreteTypes
```
Smaug123, line 320: This is the new bit, pulled up from further below.

PR #270, opened 2026-04-25
Implement ReadOnlySpan.get_Item

Smaug123:


WoofWare.PawPrint/Intrinsics.fs
```diff
@@ -1311,6 +1316,62 @@ module Intrinsics =
...
+                        failwith $"ReadOnlySpan<T>.get_Item receiver byref read produced non-value-type %O{other}"
+                | EvalStackValue.UserDefinedValueType vt -> vt
+                | other -> failwith $"ReadOnlySpan<T>.get_Item expected span receiver byref, got %O{other}"
+
+            let length : int =
+                match CliValueType.DereferenceField "_length" span |> CliType.unwrapPrimitiveLike with
+                | CliType.Numeric (CliNumericType.Int32 i) -> i
+                | other -> failwith $"ReadOnlySpan<T>.get_Item expected _length to be int32, got %O{other}"
+
+            if uint32<int32> index >= uint32<int32> length then
```
Smaug123, line 1364: we should probably also check they're nonnegative?

Smaug123: .NET doesn't; they should have been created nonnegative.

PR #290, opened 2026-04-26
Switch to using field handles, not names

### LINKED ISSUE #254: Migrate off named fields
https://github.com/Smaug123/WoofWare.PawPrint/pull/253 converted a silent failure into a host crash, but we should really never be using named fields in IDs at all. We should always find a field handle instead.

Smaug123:
Fixes #254


WoofWare.PawPrint/TypeHandleRegistry.fs
```diff
@@ -35,45 +35,31 @@ module TypeHandleRegistry =
...
-                    Type = AllConcreteTypes.getRequiredNonGenericHandle allConcreteTypes corelib.IntPtr
-                }
-                {
-                    Id = FieldId.named "m_handle"
-                    Name = "m_handle"
-                    Contents = CliType.Numeric (CliNumericType.NativeInt (NativeIntSource.TypeHandlePtr def))
-                    Offset = None
-                    Type = AllConcreteTypes.getRequiredNonGenericHandle allConcreteTypes corelib.IntPtr
-                }
-                // This is the const -1, apparently?!
```
Smaug123, line 62: > That is a const: metadata has a literal static field entry, but there is no per-object storage slot for it.

PR #303, opened 2026-04-26
Classes can be intrinsic too

Smaug123:
Sigh


WoofWare.PawPrint/Native/NativeMetadataImport.fs
```diff
@@ -127,6 +131,54 @@ module NativeMetadataImport =
 
             (state, WhatWeDid.Executed) |> ExecutionResult.Stepped |> Some
         | "System.Private.CoreLib",
+          "System.Reflection",
+          "MetadataImport",
+          "<Enum>g____PInvoke|8_0",
```
Smaug123, line 136: ugh we should fix this

PR #304, opened 2026-04-26
Implement Localloc

Smaug123:


Smaug123:
Unsuitable to go in, I think. It's not just byte arrays that can be stackalloced!

PR #306, opened 2026-04-26
Model P/Invoke last error

Smaug123:


WoofWare.PawPrint/Native/NativeKernel32.fs
```diff
@@ -0,0 +1,35 @@
+namespace WoofWare.PawPrint
+
+[<RequireQualifiedAccess>]
+module NativeKernel32 =
+    let private errorEnvVarNotFound : int = 203
```
Smaug123, line 5: how magic - see https://learn.microsoft.com/en-us/windows/win32/debug/system-error-codes--0-499- , this is ERROR_ENVVAR_NOT_FOUND

PR #324, opened 2026-04-28
Normalise managed pointer byte offsets centrally

Smaug123:
Tighten up the types.

### DIFF
--- a/WoofWare.PawPrint/BasicCliType.fs
+++ b/WoofWare.PawPrint/BasicCliType.fs
@@ -189,8 +189,58 @@ type ManagedPointerSource =
 
             projs |> List.fold formatProj rootStr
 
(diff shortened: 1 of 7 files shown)


WoofWare.PawPrint/BasicCliType.fs
```diff
@@ -189,6 +189,37 @@ type ManagedPointerSource =
 
             projs |> List.fold formatProj rootStr
 
+/// State-dependent information needed to canonicalise byte cursors.
+/// Array element size comes from the current heap; string and localloc strides
+/// are fixed by the pointer root itself.
+type ByteOffsetNormalisationContext =
+    private
+        {
+            ArrayElementSize : ManagedHeapAddress -> int
```
Smaug123: eww we should totally not have this be a function

PR #342, opened 2026-04-30
Implement RawData boxed value byte view

Smaug123:


WoofWare.PawPrint/UnaryMetadataIlOp.fs
```diff
@@ -89,7 +89,7 @@ module internal UnaryMetadataIlOp =
                             baseClassTypes
                             thread
                             activeAssy
-                            currentMethod.DeclaringType.Generics
+                            ImmutableArray.Empty
```
Smaug123, line 92: > Separate subtlety: once the method-spec arguments are concretized into methodGenerics, those are passed to resolveMember for the target generic method. But a bare MemberReference has no method-spec type arguments, so that path correctly passes ImmutableArray.Empty, not the caller’s generics.

PR #355, opened 2026-05-01
Refuse to multiply managed pointers

Smaug123:


Smaug123:
Why did Codex want to do that, it's flatly a regression

PR #379, opened 2026-05-03
Express managed byte-level byref logic

Smaug123:


WoofWare.PawPrint/IlMachineManagedByref.fs
```diff
@@ -50,6 +50,19 @@ module IlMachineManagedByref =
         | ValueSome v, ValueSome t -> v = t
         | _ -> false
 
+    let private bytesEqual (left : byte[]) (right : byte[]) : bool =
```
Smaug123, line 53: Wonder why we've got this when `=` just works

PR #381, opened 2026-05-03
Preserve managed address provenance through conv.u8

Smaug123:


WoofWare.PawPrint/ManagedPointerSource.fs
```diff
@@ -97,6 +97,103 @@ module ByteStorageIdentity =
...
+    /// Verbatim int64 bits with no managed-pointer provenance: the ordinary
+    /// signed-scalar path used by `Conv_I8`, `EvalStackValue.ofInt64`, and
+    /// arithmetic on values without address provenance.
+    | Verbatim of bits : int64
+    /// Raw address bits that originated from a managed pointer crossing
+    /// `Conv_U8` (or the storage-free normalisation of a `ManagedAddress`),
+    /// or that propagated through arithmetic with such a value. Provenance
+    /// has been flattened to bits, but the value is still semantically an
+    /// address, so signed comparisons (`clt`/`cgt`) refuse it.
+    | RawAddressBits of bits : int64
```
Smaug123, line 122: It's not clear to me why we would need this at all. Why is it not a ManagedAddress?

EDIT: Ah, I think I see. For example, a pointer offset in an array obtained by arithmetic.

PR #435, opened 2026-05-09
Implement Type.GenericParameterAttributes for type-generic parameters

Smaug123:


Smaug123:
Noooo, not the "forced intrinsics" hack

PR #457, opened 2026-05-10
Add instructions to sync the .NET runtime

Smaug123:


.claude/commands/sync-dotnet-runtime.md
```diff
@@ -0,0 +1,34 @@
...
+   `git fetch --tags` may report `! [rejected] v9.0.0-preview.6.24327.7 (would clobber existing tag)` — that's harmless and does not stop the new tags being fetched.
+
+3. Check the commit out:
+   - **If `git cat-file -e <full-sha>^{commit}` succeeds**, check out that exact commit:
+     ```bash
+     git checkout <full-sha>
+     ```
+   - **Otherwise** the version is an in-flight servicing release whose commit hasn't been pushed to `dotnet/runtime` yet (Microsoft cuts these from internal branches before the public push). Fall back to the closest published tag less than or equal to the runtime version — list candidates with `git tag --list 'v10.0.*' | sort -V` and pick the highest one that's `≤` our runtime version. If even that's missing what you need, use `origin/release/10.0`.
+
+4. Verify the source you care about compiles in your head: e.g. for QCall work, glance at `src/coreclr/System.Private.CoreLib/src/System/RuntimeHandles.cs` and `src/coreclr/vm/runtimehandles.cpp` to confirm they look like the .NET 10 shape you expect.
```
Smaug123:
```suggestion
4. Verify the source you care about compiles in your head: e.g. for QCall work against RuntimeHandles, glance at `src/coreclr/System.Private.CoreLib/src/System/RuntimeHandles.cs` and `src/coreclr/vm/runtimehandles.cpp` to confirm they look like the .NET 10 shape you expect.
```

AGENTS.md
```diff
@@ -7,7 +7,7 @@ WoofWare.PawPrint is an experimental .NET runtime implementation written in F#.
 
 This is NOT a high-performance runtime - it's a very slow IL interpreter prioritizing determinism over speed.
 
-You should find the genuine .NET runtime's source checked out at ../dotnet if you need to check behaviour; shout if it's not.
+You should find the genuine .NET runtime's source checked out at ../dotnet-runtime if you need to check behaviour; shout if it's not. The checkout should be at the .NET 10 servicing tag closest to the runtime version we use (`nix develop -c dotnet --info` reports it under `Host:`; if no exact tag exists yet — e.g. an in-flight servicing release — use the most recent `v10.0.x` tag, falling back to `origin/release/10.0` only when a needed change is unavailable in any tag). `../dotnet` (without the `-runtime` suffix) is the .NET SDK source and is not what you want.
```
Smaug123:
```suggestion
You should find the genuine .NET runtime's source checked out at ../dotnet-runtime if you need to check behaviour; shout if it's not. The checkout should be at the .NET 10 servicing tag closest to the runtime version we use (see the `.claude/commands/sync-dotnet-runtime.md` Claude command for how to make this happen). `../dotnet` (without the `-runtime` suffix) is the .NET SDK source and is not what you want.
```

PR #463, opened 2026-05-10
Implement String::FastAllocateString(MethodTable*, nint) InternalCall

Smaug123:
## Summary
- Add the .NET 10 InternalCall overload `String::FastAllocateString(MethodTable*, nint)` to `Native/NativeString.fs`. The legacy `(int)` form is now a managed wrapper that calls this one with `TypeHandleOf<string>().AsMethodTable()`. The MethodTable is strict-validated as `System.String` so we fail loudly if anything else surfaces, and we accept the broader nint encodings (`NativeInt.Verbatim`, `Int64.Verbatim`, `Int32`) the rest of CoreLib already produces. Adds `CharToStringContent.cs` as a regression test exercising `'x'.ToString()[0]`.
- Eliminate `String._firstChar` as duplicated storage. Previously char 0 lived in three places (the `_firstChar` field map entry, `StringArrayData[dataOffset]`, and `StringContents`) with manual sync points; CoreLib's `String.CreateFromChar` exposed this by initialising via `result._firstChar = c`, which the indexer/byref/span paths did not see. The fix makes `_firstChar` a synthetic projection over `StringArrayData`: `ldfld`/`ldflda`/`stfld _firstChar` all route through `getStringChar 0` / `setStringChar 0` so there is one canonical storage location, and the field is no longer materialized in the heap object's field map at allocation time.
- Generalize the field-projection seam: `RuntimeFieldProjection` now exposes `tryProjectFieldLoad` and `tryProjectFieldStore` siblings to the existing `tryProjectFieldAddress`, and `executeLdfld` / `executeStfld` consult them uniformly. This subsumes the previous ad-hoc `RawArrayDataProjection` call in `ldfld` and gives future synthetic fields one place to plug in.

## Test plan
- [x] `nix develop -c dotnet test WoofWare.PawPrint.Test/WoofWare.PawPrint.Test.fsproj` — all green
- [x] `codex review --base main` — no actionable correctness issues
- [x] `nix develop -c dotnet fantomas .`

### DIFF
--- a/WoofWare.PawPrint/UnaryMetadataFieldOps.fs
+++ b/WoofWare.PawPrint/UnaryMetadataFieldOps.fs
@@ -96,16 +96,42 @@ module internal UnaryMetadataFieldOps =
                 match state.ManagedHeap.NonArrayObjects.TryGetValue addr with
                 | false, _ -> failwith $"todo: array {addr}"
                 | true, v ->
-                    let v = AllocatedNonArrayObject.SetFieldById fieldId valueToStore v
+                    // `String._firstChar` is mirrored across three locations: the
+                    // object field, the `StringArrayData` byte view (used for byref
+                    // and span reads, and for the `string[i]` indexer's pointer
+                    // arithmetic), and the canonical `StringContents` value (used by
+                    // `String.Equals` and similar). CoreLib initialises freshly
+                    // allocated strings through this exact stfld, e.g.
+                    // `String.CreateFromChar`'s `result._firstChar = c`, so route
+                    // these writes through `setStringChar` to keep all three views
+                    // consistent. Without this, `'x'.ToString()[0]` reads the
+                    // unsynchronised side-table NUL instead of the written char.
+                    let isStringFirstChar =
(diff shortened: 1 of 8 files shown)


WoofWare.PawPrint/UnaryMetadataFieldOps.fs
```diff
@@ -96,16 +96,42 @@ module internal UnaryMetadataFieldOps =
...
+                    // object field, the `StringArrayData` byte view (used for byref
+                    // and span reads, and for the `string[i]` indexer's pointer
+                    // arithmetic), and the canonical `StringContents` value (used by
+                    // `String.Equals` and similar). CoreLib initialises freshly
+                    // allocated strings through this exact stfld, e.g.
+                    // `String.CreateFromChar`'s `result._firstChar = c`, so route
+                    // these writes through `setStringChar` to keep all three views
+                    // consistent. Without this, `'x'.ToString()[0]` reads the
+                    // unsynchronised side-table NUL instead of the written char.
+                    let isStringFirstChar =
```
Smaug123: This is mighty weird

PR #500, opened 2026-05-11
Lazily intern CastHelpers::s_table sentinel cast cache

Smaug123:
`System.Runtime.CompilerServices.CastHelpers::s_table` is a Category-B runtime-initialised static: CoreCLR's native EE populates it during `SystemDomain::LoadBaseSystemClasses` with a 2-entry sentinel int[], shaped so managed `CastCache.TryGet` always returns `MaybeCast`. PawPrint has no EE startup hook, so the field reads as null. The BCL's `CastCache.TableData` then does `ldflda RawData::Data` against null, which raises a spurious NRE; constructing its resource string re-enters SR's recursion guard and `FailFast`s with the now-familiar "Encountered infinite recursion while looking up resource 'Arg_NullReferenceException'" cascade.

Mirror the `System.String::Empty` precedent (commit 6b8bf34) by detecting the field at ldsfld/ldsflda time and installing an int[18] sentinel populated to match what `CastCache::Initialize` writes: hashShift = 63 (LeadingZeroCount on 64-bit), tableMask = 1, victimCounter = 0.

Bundle the broader docs/runtime-initialised-statics.md catalogue (the recipe for diagnosing further Category-A/B cases) and surface FailFast through the test harness as `ExecutionResult.FailFast` so the message is visible instead of being swallowed by the generic mock.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -112,6 +112,10 @@ If you find you really do need to implement a dependency, please consider whethe
 
 For detailed guidance on type concretization, generic resolution, and common patterns in the emulated CLR type system, see .claude/commands/type-concretization.md .
 
+## Runtime-Initialised Statics
+
+For the catalogue of BCL static fields whose values come from the runtime (JIT or EE) rather than any managed `.cctor` — and the recipe for diagnosing a guest `FailFast` with *"Encountered infinite recursion while looking up resource 'Arg_NullReferenceException'"* — see `docs/runtime-initialised-statics.md`.
+
 ## Instructions for OpenAI Codex agents specifically
 
 When you've completed a change to the point where you think it can be PR'ed, please commit it.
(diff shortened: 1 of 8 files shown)


AGENTS.md
```diff
@@ -112,6 +112,10 @@ If you find you really do need to implement a dependency, please consider whethe
 
 For detailed guidance on type concretization, generic resolution, and common patterns in the emulated CLR type system, see .claude/commands/type-concretization.md .
 
+## Runtime-Initialised Statics
+
+For the catalogue of BCL static fields whose values come from the runtime (JIT or EE) rather than any managed `.cctor` — and the recipe for diagnosing a guest `FailFast` with *"Encountered infinite recursion while looking up resource 'Arg_NullReferenceException'"* — see `docs/runtime-initialised-statics.md`.
+
```
Smaug123:
```suggestion
```

Smaug123: Good lord is this an inappropriate place to put this comment

PR #585, opened 2026-05-17
Implement SystemNative_InitializeTerminalAndSignalHandling as no-op success

Smaug123:
## Summary

- Adds a `libSystem.Native` handler arm for `SystemNative_InitializeTerminalAndSignalHandling` that returns truthy without mutating kernel state, matching the WASI variant in `pal_console_wasi.c`. PawPrint models neither termios nor signals, so the only correct behaviour for this entry point is to claim success and let the BCL proceed past `ConsolePal.EnsureInitializedCore`.
- The handler arm carries a comment that names the future pivot: when the first per-signal registration arm lands (`RegisterForCtrlC`, `SetTerminalInvalidationHandler`, `PosixSignalRegistration.Register`, …), it will introduce `EmulatedKernel.SignalState` and flip an `Initialized` bit that the registration arms assert. Until then the kernel is not grown speculatively.
- Adds `sourcesPure/SystemNativeInitializeTerminalAndSignalHandling.cs`: a focused `DllImport` test that exercises the entry point directly. Asserts only the invariant that holds on every supported host — a non-zero return on the first and second call — so it passes on both PawPrint and the real runtime.
- Refreshes the `WriteLine.cs` entry in `TestImpureCases.unimplemented`: the chain now reaches `SystemNative_IsATty` (called from `Console.IsOutputRedirected`), which is the freshly observed next blocker. The previous comment named `GlobalizationNative_LoadICU`/CultureInfo; that path is now bypassed since invariant culture is set unconditionally.

## Test plan

- [x] `nix develop -c dotnet fantomas .` — clean
- [x] `nix develop -c dotnet test --filter "Name~SystemNativeInitializeTerminal"` — new test passes on PawPrint and the real runtime
- [x] Regression set: `SystemNativeWrite`, `SystemNativeClose`, `SystemNativeDup`, `SystemNativeWriteSuccess` — all pass
- [x] Full `TestImpureCases` module — all pass; `WriteLine.cs` correctly remains `[<Explicit>]` via the `unimplemented` slot
- [x] `codex review --base main` — approved, no findings

### DIFF
--- a/WoofWare.PawPrint.Test/TestImpureCases.fs
+++ b/WoofWare.PawPrint.Test/TestImpureCases.fs
@@ -18,12 +18,15 @@ module TestImpureCases =
     let unimplemented =
         [
             // `Console.WriteLine("Hello, world!")` triggers lazy initialisation of `Console.Out`,
-            // which descends Console::get_Out → ConsolePal::OpenStandardOutput → Interop+Sys::Dup.
-            // PawPrint now intercepts SystemNative_Dup via FileDescriptorRegistry and
-            // SystemNative_Write via the same handler family, so both the std-stream open path and
-            // the actual byte-emitting call are implemented; the WriteLine flow now blocks
-            // downstream on the unimplemented libSystem.Globalization.Native P/Invoke
-            // `GlobalizationNative_LoadICU` during CultureInfo initialisation.
+            // which descends Console::get_Out → ConsolePal::OpenStandardOutput → Interop+Sys::Dup,
+            // then ConsolePal::EnsureInitializedCore → Interop.Sys.InitializeTerminalAndSignalHandling.
+            // PawPrint now intercepts SystemNative_Dup via FileDescriptorRegistry, SystemNative_Write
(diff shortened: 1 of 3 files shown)


Smaug123:
Closing in favour of pivoting to build the deterministic signal model before any signal-adjacent stub lands. After offline discussion we decided that the "stub-and-store" path (silently treat per-signal registration calls as no-ops) compounds badly as more Console/Process paths get unblocked, and that PawPrint's deterministic-simulation pitch wants real, harness-injectable signal delivery rather than silently-ignored signals.

The replacement work is being built in slices off `main`. Slice 1 (pure `SignalState` data model + property tests) will be opened as a fresh PR shortly; subsequent slices add dispatch, harness injection, guest self-raise, and per-signal handler arms. `SystemNative_InitializeTerminalAndSignalHandling` returns in the slice that introduces the `SignalState.Initialized` precondition; until then it stays at `failUnimplemented`, which is the same behaviour the runtime had before this PR.

PR #589, opened 2026-05-17
Support multiple handlers per signal in SignalState

Smaug123:
## Summary

- Refactor `SignalState.Registrations` from `Map<Signal, SignalHandler>` to `Map<Signal, SignalHandler list>` in registration order, so two `register` calls install sibling handlers instead of one silently replacing the other. The empty list is never a permitted value — when the last handler for a signal is unregistered, the key is dropped, so absent/empty states stay structurally equal (required for `EmulatedKernel` dedup).
- Mirrors real .NET: each `PosixSignalRegistration.Create` adds a `Token` to a per-signo list; dispatch iterates in reverse registration order with `Cancel`-based short-circuit. The dispatcher (a later slice) is responsible for the reversal; `SignalState` stores registration order.
- API change: `tryGetHandler` → `handlers` (returns the registration-order list, empty when none). `unregister` now takes the `SignalHandler` to remove and drops the first structural match only. `tryDeliverable` returns the full handler list.
- 5 new unit tests for the multi-handler edges (append order, duplicate preservation, no-op on non-match, first-match-only removal, empty-list key drop). The property test gains three distribution counters confirming the random walk actually reaches multi-handler states (rather than degenerating to the single-handler case).

## Test plan

- [x] `nix develop -c dotnet test --filter "FullyQualifiedName~TestSignalState"` — 31/31 pass (property test runs 500 random sequences).
- [x] `nix develop -c dotnet test WoofWare.PawPrint.Test/WoofWare.PawPrint.Test.fsproj` — 1198/1198 pass.
- [x] `nix develop -c dotnet fantomas .` — clean.
- [x] `codex review --base main` — no actionable findings.

### DIFF
--- a/WoofWare.PawPrint.Test/TestSignalState.fs
+++ b/WoofWare.PawPrint.Test/TestSignalState.fs
@@ -61,7 +61,7 @@ module TestSignalState =
     let ``empty has nothing registered, nothing blocked, nothing pending`` () : unit =
         let s = SignalState.empty
         SignalState.isInitialized s |> shouldEqual false
-        SignalState.tryGetHandler Signal.SIGINT s |> shouldEqual None
+        SignalState.handlers Signal.SIGINT s |> shouldEqual []
         SignalState.isBlocked t0 Signal.SIGINT s |> shouldEqual false
         SignalState.blockedFor t0 s |> shouldEqual Set.empty
         SignalState.pending s |> Seq.toList |> shouldEqual []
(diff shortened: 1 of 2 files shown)


Smaug123:
Closing as throwaway. On further reflection, the multi-handler shape for `SignalState.Registrations` was an over-eager abstraction: the P/Invoke arms PawPrint will implement (`SystemNative_EnablePosixSignalHandling` etc.) carry only signo/enable bits, not handler info, so `Registrations` has no honest path to get populated from PawPrint's kernel-side seam. The handler dictionary belongs in the BCL's simulated managed heap (which PawPrint mutates by executing `PosixSignalRegistration.Create` as managed code), and the eventual dispatcher will invoke `PosixSignalRegistration.OnPosixSignal` on a chosen thread rather than walking handlers from kernel state. The next slice will collapse `Registrations` into `Enabled : Set<Signal>` (mirroring the C-side enable bits) and add the three P/Invoke arms with their now-clear semantics.

PR #594, opened 2026-05-17
Implement MarshalNative_TryGetStructMarshalStub for strictly-numeric structs

Smaug123:
## Summary
- Implements the `MarshalNative_TryGetStructMarshalStub` QCall, but only for the narrow strictly-numeric-blittable subset (recursive structs whose leaves are `Int8`..`Float64`, excluding `NativeInt`).
- For that subset, writes a null stub + the managed CLI size and returns TRUE so `Marshal.StructureToPtr` / `Marshal.PtrToStructure` can take the `SpanHelpers.Memmove` fast path.
- Anything else (IntPtr/UIntPtr fields, enums, `[MarshalAs]` descriptors, Bool/Char/ObjectRef fields, AutoLayout types, classes) surfaces a host TODO. Each future widening is left to its own focused PR with a motivating PawPrint test.

## Why minimal
A previous attempt grew into a full reimplementation of CoreCLR's `IsFieldBlittable` plus stub-synthesis exception modelling plus byte-copyability gating plus empty-struct size bump plus CharSet handling plus primitive-like unwrap. Each codex round found new gaps in that classifier because each gap is a separate piece of CoreCLR behaviour. Shrinking back to the strictly-numeric allowlist keeps the surface honest: we say TRUE exactly when we're confident, and TODO everywhere else.

The fuller attempt is preserved on `marshal-try-get-struct-stub-full-attempt` for reference, but is not the intended trajectory.

## Effect
`AdvancedStructLayout.cs` advances past this QCall (its only `Marshal.StructureToPtr` call uses a strictly-numeric `BlittableStruct`). It stays in `unimplemented` because `Unsafe.ByteOffset` does not yet support the `Pointer(...)` `ManagedPointerSource` produced when `Marshal.StructureToPtr` forms `&*(byte*)ptr` over an `AllocHGlobal` pointer.

## Follow-ups (separate PRs / issues)
- Return FALSE for no-layout types (`new object()` etc.) so `Marshal` can throw `ArgumentException`.
- Stub-synthesis-style TypeLoadException/MarshalDirectiveException for has-layout-non-blittable types.
- IntPtr/UIntPtr field handling, with byte-copyability provenance gating.
- Enum / primitive-like value-type fields.
- `[MarshalAs]` descriptor handling (ByValTStr, ByValArray, scalar width matching, Struct).
- Bool/Char fields, with declaring-type CharSet.
- Empty-struct native-size bump to 1 byte.
- Decimal-as-field rejection (Decimal's own MethodTable is blittable but it forces stub synthesis as a field).

## Test plan
- [x] `dotnet build` clean
- [x] Full `dotnet test` suite passes (1162 passed, 0 failed)
- [x] `dotnet fantomas` reports no formatting changes
- [x] `AdvancedStructLayout.cs` test remains in `unimplemented` with a refreshed reason

### DIFF
--- a/WoofWare.PawPrint.Test/TestPureCases.fs
+++ b/WoofWare.PawPrint.Test/TestPureCases.fs
@@ -17,7 +17,7 @@ module TestPureCases =
 
     let unimplemented =
         [
-            "AdvancedStructLayout.cs" // past MarshalNative_SizeOfHelper for ByValTStr and SystemNative_Malloc / SystemNative_Free / Marshal.AllocHGlobal / FreeHGlobal; now blocked downstream at the unimplemented MarshalNative_TryGetStructMarshalStub QCall (CoreLib's Marshal.StructureToPtr path)
+            "AdvancedStructLayout.cs" // past MarshalNative_TryGetStructMarshalStub for the strictly-numeric blittable struct (Marshal.StructureToPtr now hits the SpanHelpers.Memmove fast path); now blocked downstream in Unsafe.ByteOffset, which doesn't yet support a `Pointer(...)` ManagedPointerSource (one of the byrefs comes from `&*(byte*)ptr` over an AllocHGlobal IntPtr)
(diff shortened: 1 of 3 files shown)


Smaug123:
Tracking issue for the deferred classifier work: #595

PR #639, opened 2026-05-20
Implement XplatEventLogger QCalls deterministically

Smaug123:
## Summary

Implements the three `XplatEventLogger` QCalls (`EventSource_GetClrConfig`, `IsEventSourceLoggingEnabled`, `LogEventSource`) directly in a new `NativeEventSource` module. These QCalls are only present in CoreLibs built with `FEATURE_EVENTSOURCE_XPLAT` (the Linux build), but PawPrint always dispatches against the host runtime's CoreLib, so they must be registered unconditionally.

The handlers honour the guest's emulated environment rather than no-opping:

- **`EventSource_GetClrConfig(name)`** looks up `DOTNET_<name>` (falling back to `COMPlus_<name>`) in `state.Kernel.Environment`, encodes the value as a freshly-allocated UTF-16 buffer with a NUL terminator, and returns the pointer. Unset/empty values yield a null pointer (CoreLib's `new string((char*)null)` then collapses to `String.Empty`), matching CoreCLR's `GetConfigString` and `EnvGetString` behaviour (`clrconfig.cpp:131`, `:288`).

- **`IsEventSourceLoggingEnabled()`** parses `DOTNET_EnableEventLog` as a CLRConfig DWORD and returns `Int32 1` iff the parsed value is non-zero. Defaults to disabled when the knob is unset or unparseable. This faithfully mirrors `XplatEventLogger::IsEventLoggingEnabled()` (`eventtracebase.h:489`) — when it returns false, `InitializePersistentListener` short-circuits and `LogEventSource` is unreachable. The C# wrapper carries `[return: MarshalAs(UnmanagedType.Bool)]`, so the underlying QCall is `int32`-returning; the wrapper converts via `cgt.un`.

- **`LogEventSource(...)`** fails loud. It is only reachable when the guest enables tracing via `DOTNET_EnableEventLog`, and PawPrint has no LTTng / EventPipe consumer to forward events to. Silently dropping the event would lose data the guest asked us to surface; the failwith points at the opt-out knob (`DOTNET_EnableEventLog=0`) for guests that don't actually care about the trace stream.

### `tryParseClrConfigDword`: matching CoreCLR's PAL_wcstoul

`EnableEventLog` is declared via `RETAIL_CONFIG_DWORD_INFO` with no `ParseIntegerAsBase10` flag (`clrconfigvalues.h:580`), so `GetConfigDWORD` (`clrconfig.cpp:228`) parses with radix 16. The implementation mirrors the Unix PAL's `PAL_wcstoul` (`pal/src/cruntime/wchar.cpp:281-324`), which wraps glibc `strtoul` on a 64-bit host (`unsigned long` is 64-bit):

- Skip leading whitespace; accept an optional `+` / `-` sign.
- Accept an optional `0x` / `0X` radix prefix, but only when at least one hex digit follows it (otherwise the leading `0` is itself the parsed digit and `x` is the stop character).
- Consume the longest hex-digit prefix and ignore everything after it (so `1garbage` parses as `1`).
- Parse the magnitude as `uint64`, then apply the sign in 64-bit arithmetic and truncate to `uint32`. This captures the PAL's HOST_64BIT special case: a negative input whose magnitude exceeds `UINT32_MAX` is **not** rejected — `strtoul` parses it in 64-bit, two's-complement wraps the negation, and the PAL's final `(ULONG)res` cast truncates to the low 32 bits without setting `ERANGE`. So `DOTNET_EnableEventLog=-100000001` enables logging on real CoreCLR (low 32 bits = `0xFFFFFFFF`).
- Return `None` only in CoreCLR's two failure arms: `endPtr == val` (no digits consumed) or `errno == ERANGE` (magnitude exceeded `uint64`, or positive magnitude exceeded `UINT32_MAX` — the PAL's HOST_64BIT clamp-and-set-ERANGE arm). The caller treats `None` as the default `0`, i.e. disabled.

### Files

- `WoofWare.PawPrint/Native/NativeEventSource.fs` — new module with the three handlers, the parser, and the env-lookup helper.
- `WoofWare.PawPrint/Native/NativeQCall.fs` — registers the three handlers.
- `WoofWare.PawPrint/WoofWare.PawPrint.fsproj` — adds the new file.
- `WoofWare.PawPrint.Test/TestNativeEventSource.fs` — 24 direct tests (see below).
- `WoofWare.PawPrint.Test/WoofWare.PawPrint.Test.fsproj` — adds the new test file.

## Test plan

The macOS CoreLib does not contain `XplatEventLogger` (it's `FEATURE_EVENTSOURCE_XPLAT`-gated for Linux), so we can't drive the new handlers through a normal end-to-end IL fixture on the local dev box. Instead, the tests build a synthetic `NativeCallContext` directly: a real concretized donor method (`Thread.YieldInternal`) has its `Signature` overridden to the shape each test needs, and a record-updated `TypeInfo` provides the `XplatEventLogger` namespace/name. The dispatcher reads neither the donor identity nor the donor declaring type, so this is sound.

- [x] **Pure helper tests** (10):
    - `tryParseClrConfigDword`: empty/whitespace → None; hex by default; `0x` prefix tolerated; longest-prefix parsing (`1garbage` → 1, `ff thanks` → 0xff, `0XYZ` → 0); `0x` without trailing hex parses the leading `0`; sign with two's-complement wraparound (`+1` → 1, `-1` → 0xFFFFFFFF, `-0` → 0); leading whitespace skipped; positive overflow past UINT32_MAX → None (PAL HOST_64BIT ERANGE arm); negative overflow past UINT32_MAX truncates to low 32 bits (Codex-flagged case: `-100000001` → 0xFFFFFFFF, `-100000000` → 0, `-1FFFFFFFF` → 1); no-digits → None.
    - `lookupClrConfigString`: `DOTNET_` wins over `COMPlus_`; `COMPlus_` fallback when `DOTNET_` unset; empty value treated as unset (mirrors CoreCLR's `*ret != W('\0')` filter); nothing set → None.

- [x] **Dispatcher tests for `IsEventSourceLoggingEnabled`** (9): unset → 0; `=0` → 0; `=1` → 1; hex `10` (= 16) → 1; `COMPlus_EnableEventLog=1` wired up; malformed value with no parseable digits → 0; `=-1` enables (wcstoul wraps to `0xFFFFFFFF`); `=1garbage` enables (longest-prefix parse); `=-100000001` enables (PAL_wcstoul 64-bit wrap — Codex-flagged regression).

- [x] **Dispatcher test for `LogEventSource`** (1): synthesizes the 4-arg signature and asserts the failwith fires with a message that points at the `DOTNET_EnableEventLog=0` opt-out.

- [x] Full local test suite passes: 1376 tests on macOS (`nix develop -c dotnet test --configuration Release`).
- [x] Two rounds of `codex review --base main`; final review returned no actionable findings.
- [ ] Linux CI exercises the new handlers through CoreLib's `XplatEventLogger` callers as well.

### DIFF
--- a/WoofWare.PawPrint/Native/NativeEventSource.fs
+++ b/WoofWare.PawPrint/Native/NativeEventSource.fs
@@ -2,28 +2,106 @@ namespace WoofWare.PawPrint
 
 [<RequireQualifiedAccess>]
 module NativeEventSource =
-    /// QCalls declared on `System.Diagnostics.Tracing.XplatEventLogger` and reached
-    /// from CoreLib only when `FEATURE_EVENTSOURCE_XPLAT` was defined at CoreLib
-    /// build time (i.e. the Linux-built `System.Private.CoreLib`). They are still
-    /// registered unconditionally because PawPrint always dispatches against the
-    /// host runtime's CoreLib, which can be the Linux one on a Linux dev box or
-    /// CI runner.
+    /// Parse a CLRConfig DWORD env-var value the way CoreCLR does for
+    /// `EnableEventLog`: `wcstoul(val, &endPtr, 16)` — hex by default, with the
+    /// optional `0x` prefix tolerated. Returns `None` if the value is empty or
+    /// fails to parse (CoreCLR's `GetConfigDWORD` falls back to the default
+    /// value, which is `0` for `EnableEventLog`).
     ///
-    /// PawPrint does not connect to any external tracing consumer (LTTng,
-    /// EventPipe, etc.), so the deterministic answers are:
-    ///   * `EventSource_GetClrConfig`: report "no config set" by returning a
-    ///     null UTF-16 pointer. CoreLib's `new string((char*)null)` then yields
-    ///     `String.Empty`, matching real CoreCLR when the named config knob is
-    ///     unset (see NativeString.fs / `String..ctor(char*)`).
-    ///   * `IsEventSourceLoggingEnabled`: return `FALSE` so
-    ///     `XplatEventLogger.InitializePersistentListener` skips creating the
-    ///     listener entirely. This is the truthful answer in a host with no
-    ///     tracing consumer attached.
-    ///   * `LogEventSource`: void no-op. With `IsEventSourceLoggingEnabled`
-    ///     returning false the listener is never created and this entry point
-    ///     should be unreachable from `OnEventWritten`, but implementing it
-    ///     defensively keeps PawPrint robust against guests that import it
-    ///     directly.
+    /// `EnableEventLog` is declared via `RETAIL_CONFIG_DWORD_INFO` with no
+    /// `ParseIntegerAsBase10` flag (see `clrconfigvalues.h:580`), so the radix
+    /// is 16. Both `1` and `0x1` therefore mean TRUE; `10` means 16 (also
+    /// TRUE); `0` and `0x0` mean FALSE. We don't need the full DWORD value —
+    /// `IsEventSourceLoggingEnabled` only asks whether it's non-zero — but
+    /// the parser is shaped to surface the value so future knobs that care
+    /// about the numeric magnitude can reuse it.
+    ///
+    /// Exposed as `internal` so the test assembly can pin the parser's
+    /// behaviour directly without going through the QCall dispatcher.
+    let internal tryParseClrConfigDword (raw : string) : uint32 option =
+        let trimmed = raw.Trim ()
+
+        if System.String.IsNullOrEmpty trimmed then
+            None
+        else
+            let hexBody =
+                if trimmed.StartsWith ("0x", System.StringComparison.OrdinalIgnoreCase) then
+                    trimmed.Substring 2
+                else
+                    trimmed
+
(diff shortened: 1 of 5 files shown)


WoofWare.PawPrint/Native/NativeEventSource.fs
```diff
@@ -0,0 +1,81 @@
...
+        with
+        | "EventSource_GetClrConfig",
+          "System.Private.CoreLib",
+          "System.Diagnostics.Tracing",
+          "XplatEventLogger",
+          [ ConcretePointer (ConcretePrimitive state.ConcreteTypes PrimitiveType.UInt16) ],
+          MethodReturnType.Returns (ConcretePointer (ConcretePrimitive state.ConcreteTypes PrimitiveType.Char)) ->
+            // The single argument is the UTF-16 config-name pointer (marshalled
+            // from the C# `string` parameter); intentionally not read because
+            // PawPrint has no configured EventSource knobs.
```
Smaug123: Ugh this seems exactly like the kind of silent failure we don't want

PR #640, opened 2026-05-20
Implement PCT scheduler policy with per-step weighted demotion

Smaug123:
## Summary

Lights up the schedule-fuzzing PCT (Probabilistic Concurrency Testing) policy on the `SchedulerState` plumbing landed in #638. With this PR, running `WoofWare.PawPrint.App --seed <N> ...` selects PCT with the given seed; without a seed, the runtime keeps its existing round-robin behaviour. Same seed ⇒ same observable outcome.

## Algorithm

Per-step weighted-Bernoulli demotion (rather than Burckhardt's original fixed-`d` schedule):

1. Lazily sample a uniform-on-`[0, 1)` priority for each Runnable thread on first observation.
2. Pick the argmax priority.
3. Run one Bernoulli draw against `weight * P_BASE` (P_BASE = 0.01, weight from `ContextSwitchPrior` — 1.0 for native frames and `AlwaysGuestVisible` ops, down to 0.0 for `Never`).
4. On a hit, resample the running thread's priority (likely demoting it — may accidentally re-promote, by design).

Sampling is lazy rather than eager at thread-creation sites so determinism can't be silently broken by a future thread-spawn path forgetting to call the scheduler. Argmax over the sorted `runnable` list gives deterministic tie-breaking (lowest `ThreadId` wins — astronomically rare with 53-bit mantissas, but pinned anyway).

## Wiring

- `IlMachineState.withPctSeed` installs a Pct policy from a `uint64` seed.
- `Program.prepare`/`Program.run` take a `pctSeed : uint64 option`; `None` keeps the default RoundRobin.
- `Scheduler.onThreadTerminated` drops the terminated thread's priority entry so the map domain stays "ever-seen-Runnable-and-not-terminated".
- `WoofWare.PawPrint.App` accepts `--seed <decimal-or-0xHEX>` (in either order with `--debug-server`) and echoes the resolved seed to stderr in canonical `0x%016X` form so it's copy/pasteable into a reproducer.
- `ThreadState.peekNextOp` is a side-effect-free imminent-op lookup used only by the PCT classifier; native bodies (`InternalCall`/`PInvoke`/`RuntimeProvided`) return `None` (treated as `AlwaysGuestVisible`).

## Tests

`TestSchedulerPct.fs`:
- Pure `PctState` helpers: range, lazy insert, sampling order, removal idempotence.
- Scheduler integration (no live-frame paths): `chooseNext` on empty Runnable, `hasAnyRunnable` agreement, termination cleanup.
- End-to-end reproducibility on `ReadWriteRace.cs`: two seeded runs produce identical outcomes; `None` seed stays RoundRobin and reproduces too.

## Test plan
- [x] `nix develop -c dotnet build WoofWare.PawPrint.slnx` — clean
- [x] `nix develop -c dotnet fantomas .` — formatted
- [x] `nix develop -c dotnet test WoofWare.PawPrint.Test/...` — 1352/1352 pass
- [ ] `codex review --base main` (next)

### DIFF
--- a/WoofWare.PawPrint.App/DebuggerServer.fs
+++ b/WoofWare.PawPrint.App/DebuggerServer.fs
@@ -328,12 +328,13 @@ module DebuggerServer =
         (dotnetRuntimeDirs : ImmutableArray<string>)
         (impls : NativeImpls)
         (env : Map<string, string>)
+        (pctSeed : uint64 option)
         (argv : string list)
         : SessionState
         =
         use fileStream = new FileStream (dllPath, FileMode.Open, FileAccess.Read)
 
-        match Program.prepare loggerFactory (Some dllPath) fileStream dotnetRuntimeDirs impls env argv with
+        match Program.prepare loggerFactory (Some dllPath) fileStream dotnetRuntimeDirs impls env pctSeed argv with
         | Program.ProgramStartResult.Ready prepared -> SessionState.Running (prepared, 0L)
         | Program.ProgramStartResult.CompletedBeforeMain outcome -> SessionState.Finished (outcome, 0L)
 
@@ -973,6 +974,7 @@ module DebuggerServer =
         (dotnetRuntimeDirs : ImmutableArray<string>)
         (impls : NativeImpls)
         (env : Map<string, string>)
+        (pctSeed : uint64 option)
         (argv : string list)
         (token : string)
         (configureWebHost : IWebHostBuilder -> unit)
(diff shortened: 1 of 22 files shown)


Smaug123:
The music Claude proposes to accompany this PR is _Piano Phase_ by Steve Reich. After this PR, PawPrint is capable of identifying and exactly reproducing the flaky test https://github.com/Smaug123/WoofWare.PawPrint/blob/52194ad467c8e5c8e516fdf7012fb6ede23bba0c/WoofWare.PawPrint.Test/sources/ReadWriteRace.cs .

PR #658, opened 2026-06-01
Add EmulatedRuntime descriptor + drift-guard test

Smaug123:
## What

Introduces an explicit, queryable descriptor of *which* .NET runtime PawPrint reproduces, plus a guard against it silently going stale.

- **`EmulatedRuntime` type + `EmulatedRuntime.current`** (`net10`): records the target framework, servicing version (10.0.7), and the public `dotnet/runtime` release tag + commit (`v10.0.7` / `7706f546`). Documented as *descriptive metadata about what we aim to reproduce* — not the source of truth for behaviour, which is PawPrint's own code/version.
- **`NativeDispatch.nativeHandlers` → `net10NativeHandlers`**: names the native handler list as the net10 set and documents that a second runtime would get its own list selected on `EmulatedRuntime`. No behaviour change; no premature plumbing.
- **`TestEmulatedRuntime`**: asserts `EmulatedRuntime.current.Version` matches the runtime the suite runs on (`RuntimeInformation.FrameworkDescription`). Fails loudly when the Nix-pinned SDK drifts ahead of the constant (verified: fails at 10.0.6 against a 10.0.7 runtime).

## Why

Foundation for versioning/publishing PawPrint: the package version tracks PawPrint's own API (semver), while *which net10 we emulate* is a separate axis surfaced here rather than baked into the version number.

## Notes

`SourceCommit` is the public release-tag commit (the readable source we mirror), deliberately **not** the binary's internal `.version` build commit — that one is often not pushed publicly. See the in-code comment.


WoofWare.PawPrint/EmulatedRuntime.fs
```diff
@@ -0,0 +1,51 @@
...
+        /// pushed to the public dotnet/runtime repo. We record the public, readable source commit.
+        SourceCommit : string
+    }
+
+[<RequireQualifiedAccess>]
+module EmulatedRuntime =
+
+    /// .NET 10. The only runtime PawPrint currently emulates.
+    ///
+    /// Keep these values in step with the runtime pinned by the Nix devshell and the
```
Smaug123: this should be comments, not docstrings

PR #659, opened 2026-06-01
Pin the dotnet/runtime reference source in the flake

Smaug123:
## What

Replaces the ad-hoc `../dotnet-runtime` sibling checkout (which silently drifted from the devshell runtime) with a pinned, reproducible reference source.

- **`dotnet-runtime-src`**: a sparse `fetchgit` of `dotnet/runtime` at the public `v10.0.7` tag, exposed in the devshell as `$DOTNET_RUNTIME_SRC`. Sparse-checked-out to `src/coreclr`, `src/libraries/System.Private.CoreLib`, and `eng` to keep the closure small.
- **`runtime-version-pin` flake check** (runs in CI's `flake-check` job): fails when nixpkgs's runtime version drifts from `expectedRuntimeVersion`, forcing a deliberate bump of the pin, `expectedRuntimeVersion`, and `EmulatedRuntime.fs` together.
- **AGENTS.md + `sync-dotnet-runtime`** rewritten for the new mechanism (edit `flake.nix`, no sibling checkout).

## Provenance notes (the confusing bits, documented in-tree)

- Pin to the **public `vX.Y.Z` tag commit**, not the binary's internal `.version` build commit — the latter is frequently not pushed publicly and so can't be fetched.
- dotnet tags the release commit **before** bumping the in-tree version, so the `v10.0.7` tree self-reports `PatchVersion 6`. It is genuinely the 10.0.7 release source (`v10.0.5→4`, `v10.0.7→6`, `v10.0.8→7`). Pin by the tag, not the in-tree string.

## Depends on #658

The new docs and the `runtime-version-pin` check message reference `WoofWare.PawPrint/EmulatedRuntime.fs` and `TestEmulatedRuntime`, which land in #658. **Merge #658 first** so those references resolve on `main`. (Functionally independent — the check only depends on `dotnet-runtime.version` — so either order builds, but #658-first keeps the docs accurate.)


.claude/commands/sync-dotnet-runtime.md
```diff
@@ -1,36 +1,41 @@
...
...` enum at `$DOTNET_RUNTIME_SRC/src/coreclr/inc/corinfo.h`. As of the audit it has exactly three entries (`ZERO`, `EMPTY_STRING`, `ISLITTLEENDIAN`) and `TestBclIntrinsicStaticFields` in `WoofWare.PawPrint.Test` pins the corresponding BCL field set. If a new entry appears, the JIT is folding `ldsfld` for a slot that PawPrint reads literally: audit the BCL declaration (`grep -nR "\[Intrinsic\]" $DOTNET_RUNTIME_SRC/src/libraries/System.Private.CoreLib/src/`) for any new static field that lacks an initialiser. Don't assume that "the zero-initialised slot is the right value" makes the case safe — `IntPtr::Zero` and `UIntPtr::Zero` look that way but `cliTypeZeroOf` populates them with `NativeIntSource.ManagedPointer Null`, which compares unequal to `Verbatim 0L` in `cgt.un`. Always add a focused guest test (see `IntPtrZero.cs` / `UIntPtrZero.cs` / `BitConverterIsLittleEndian.cs`) before relying on the generic path; lazy population at first `ldsfld`/`ldsflda` (the `System.String::Empty` approach in `UnaryMetadataFieldOps.executeLdsfld`) is the safer default for non-Boolean fields. Then update `expectedIntrinsicStaticFields` in `TestBclIntrinsicStaticFields.fs` to acknowledge the new field.
```
Smaug123: Wait what

> `NativeIntSource.ManagedPointer Null`, which compares unequal to `Verbatim 0L` in `cgt.un`

PR #673, opened 2026-08-03
Implement SystemNative_GetCryptographicallySecureRandomBytes

Smaug123:
`Guid.NewGuid()` on Unix goes straight through
`Interop.Sys.GetCryptographicallySecureRandomBytes(byte*, int) -> int`, which
PawPrint did not implement — so any guest that minted a Guid died at the
P/Invoke boundary with `Unimplemented native method`.

Confirmed against the shipped CoreLib rather than from memory:

```
// Interop::static GetCryptographicallySecureRandomBytes(ptr[uint8], int32) : void
IL_0002: Call     Interop/Sys::GetCryptographicallySecureRandomBytes
IL_0007: UnaryConst.Brfalse_s 6
IL_0009: Newobj   System.Security.Cryptography.CryptographicException::.ctor
IL_000E: Throw
```

So 0 means success, and CoreLib's wrapper throws `CryptographicException` on
anything else.

## Approach

Implemented the same way as its already-supported non-crypto sibling: a seeded
splitmix64 substitute for host entropy, because bit-for-bit reproducibility is
the property this runtime exists to provide and no deterministic interpreter can
honour the "cryptographically secure" contract anyway. **Under PawPrint this
entry point is not cryptographically secure in any sense**; that is documented
at both the kernel field and the handler.

Two notes on the shape:

* The entry point reports status, unlike its sibling. PawPrint's substitute has
  no failure mode — there is no host entropy source to be exhausted or
  unreadable — so it always returns 0. Malformed arguments (negative length,
  null destination for a non-zero length) still abort loudly rather than
  masquerading as an entropy failure, since those are guest/interpreter bugs
  rather than the condition `CryptographicException` describes.
* It draws from its own `EmulatedKernel.CryptoRandomState` stream rather than
  sharing `NonCryptoRandomState`. This follows the guidance already on
  `NonCryptoRandom`: sharing one state would make an added `new Random()` (or
  any other non-crypto consumer) silently shift every subsequent
  `Guid.NewGuid`, which is exactly the kind of action at a distance that makes
  a recorded trace hard to reason about. Seeded from a distinct
  nothing-up-my-sleeve constant so the two streams do not emit identical bytes
  from a fresh kernel.

The decode/validate/fill body is now shared between the two entry points, which
declare the identical `(byte* buffer, int32 bufferLength)` argument list.

## Tests

Written before the implementation and observed failing on `Unimplemented native
method`, then passing:

* `sourcesPure/SystemNativeGetCryptographicallySecureRandomBytes.cs` — direct
  P/Invoke stub mirroring CoreLib's declaration. Runs against the real CLR and
  PawPrint alike, so it asserts only what holds of any correct implementation:
  return code 0, zero-length draw is a no-op, the sentinel tail past the
  requested length is untouched, and two successive draws differ.
* `sourcesPure/GuidNewGuid.cs` — end-to-end `Guid.NewGuid()`, checking the RFC
  4122 version-4 and variant-10xx bits. This passes, so it does not need to go
  in `unimplemented`.
* `TestNonCryptoRandom.fs` — the two kernel seeds are non-zero and distinct,
  pinning the property the separate-stream decision rests on.

Full suite: 1443/1443 passing. `codex review --base main` found nothing.

### DIFF
--- a/WoofWare.PawPrint/EmulatedKernel.fs
+++ b/WoofWare.PawPrint/EmulatedKernel.fs
@@ -387,6 +387,27 @@ type EmulatedKernel =
         /// `Random()` ctor retries until it sees a non-zero seed, so a
         /// constant-zero substitute would hang at construction time.
         NonCryptoRandomState : uint64
+        /// Deterministic state for the splitmix64 PRNG that backs
+        /// `SystemNative_GetCryptographicallySecureRandomBytes` — the entry
+        /// point `Guid.NewGuid` draws its 16 bytes from on Unix, and the one
+        /// CoreLib's `Interop.GetCryptographicallySecureRandomBytes` wrapper
+        /// turns into a `CryptographicException` on any non-zero return.
+        /// PawPrint substitutes the same seeded PRNG it uses for the
+        /// non-crypto entry point: the output is emphatically *not*
+        /// cryptographically secure, but nothing inside a deterministic
+        /// interpreter can be, and reproducibility is the property this
+        /// runtime exists to provide. Guests that need real entropy must not
+        /// run under PawPrint.
+        ///
+        /// Deliberately a *separate* stream from `NonCryptoRandomState`,
+        /// per the guidance on `NonCryptoRandom`: sharing one state would
+        /// make an added `new Random()` (or any other non-crypto consumer)
+        /// silently shift every subsequent `Guid.NewGuid`, which is exactly
+        /// the kind of spooky action at a distance that makes a recorded
+        /// trace hard to reason about. Seeded from a constant distinct from
+        /// `NonCryptoRandom.initialState` so the two streams do not emit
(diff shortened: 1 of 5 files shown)


WoofWare.PawPrint/EmulatedKernel.fs
```diff
@@ -437,6 +458,15 @@ module EmulatedKernel =
     let defaultEnvironment : Map<string, string> =
         Map.ofList [ "DOTNET_SYSTEM_GLOBALIZATION_INVARIANT", "1" ]
 
+    /// Seed for `EmulatedKernel.CryptoRandomState`. The first 64 bits of the
+    /// fractional part of pi — a nothing-up-my-sleeve constant chosen purely
```
Smaug123, line 462: whomst among us does not know this

PR #677, opened 2026-08-04
Classify [Intrinsic] against the post-resolution method

Smaug123:
Follow-up to #669, which left `ArrayCloneThroughICloneable.cs` in `unimplemented` documenting this bug.

## The bug

`IlMachineStateExecution.callMethod` computed `[Intrinsic]`-ness against the method as passed in, but performed virtual/interface resolution ~330 lines later, with the intrinsic dispatch site sitting between the two.

So a `callvirt` of a non-intrinsic abstract declaration cleared the intrinsic check (suppressed by `isAbstractBody`), then resolved to an `[Intrinsic]` override whose IL body PawPrint went on to interpret — exactly what `[Intrinsic]` means we must not do. Concretely, `callvirt ICloneable::Clone()` on an array resolved to `System.Array::Clone`, whose body reaches the unimplemented `Object.MemberwiseClone`.

Hoisting alone broke 67 tests, all one cause: `RuntimeType.get_TypeHandle`, which `callvirt Type::get_TypeHandle()` now correctly resolves to, carries its own `[Intrinsic]` with no handler.

A green suite then proved insufficient twice, so the surface was audited directly by enumerating every `[Intrinsic]` CoreLib method that virtual resolution can select (**118**). The change moves behaviour in two directions:

**Newly intercepted** (declaration non-intrinsic → override intrinsic):
- `Array::Clone` — implemented in #669.
- `RuntimeType::get_TypeHandle` — allowlisted. IL is `ldarg.0; newobj RuntimeTypeHandle::.ctor(RuntimeType); ret`, and the ctor is a single `m_type` field store.
- `String::Equals(string)` — allowlisted. **This was a real regression**: `IEquatable<string>::Equals` is abstract and non-intrinsic, so it clears the check, and resolution selects the `[Intrinsic]` instance overload, which the hand-written `String.Equals` arm rejects (it only accepts the static two-argument form). Passes on main, failed here until fixed. Found by review, not by the suite.
- Long tail (`Int128`, `Vector*`, `Matrix*`, `Quaternion`, `NFloat`, `CLong`, `CULong`). Probed rather than assumed: `Int128` already dies on an unrelated *static* intrinsic (`op_Implicit`) on main, so these types are unusable in PawPrint either way. Theoretical.

**No longer intercepted** (declaration intrinsic → override non-intrinsic): only bites where a type actually overrides an `[Intrinsic]` virtual. `get_IsEnum`, `get_IsValueType` and `get_IsGenericType` are *not* overridden by `RuntimeType` (it overrides `IsValueTypeImpl`, not the property), so those arms fire unchanged. `GetGenericTypeDefinition` *is* overridden — and pre-hoist would have hit the `failwith`, so this improves it.


WoofWare.PawPrint/IntrinsicMethodKeys.fs
```diff
@@ -153,12 +153,52 @@ module IntrinsicMethodKeys =
...
                 "GetTypeFromHandle"
                 [ IntrinsicParameterPattern.Exact "System.RuntimeTypeHandle" ]
+            // `RuntimeType.TypeHandle`'s getter overrides the [Intrinsic] `Type.TypeHandle`
+            // getter and carries its own [Intrinsic] solely "to avoid round-trip
+            // handle -> RuntimeType -> handle in JIT" (its own source comment). The IL body is
+            // `ldarg.0; newobj RuntimeTypeHandle::.ctor(RuntimeType); ret`, and that ctor is
+            // `internal RuntimeTypeHandle(RuntimeType? type) { m_type = type; }` — a single field
+            // store. Interpreting it produces the same `m_type`-populated RuntimeTypeHandle that
+            // PawPrint used to synthesise by hand, so the hand-written `Type.get_TypeHandle` arm
+            // in Intrinsics.fs was retired in favour of running the real IL.
```
Smaug123: Ugh no narrative comments please

PR #679, opened 2026-08-04
Allow `unbox.any` between an enum and its underlying primitive

Smaug123:
Resolves the standing TODO in `executeUnboxAny` (last of the `unbox.any` series, after #670, #672, #675).

`(int)(object)someEnum` raised InvalidCastException where a real runtime returns the value.

Interestingly, verification types collapse signedness — `int32` and `uint32` share one — but the runtime does not: `(uint)(object)1` throws.

Simplest test: `(int)(object)someEnum` emits `unbox.any int32` directly.

## The actual rule

CoreCLR `CastHelpers.Unbox_Helper`: accept when the handles are identical, **or** when both types are in the primitive-value-type category and report the *same* primitive element type. Enums are in that category (`// Enum is included`, RuntimeHelpers.CoreCLR.cs) and report their underlying integer's element type.

Checked against the real runtime across ~30 cases. That one rule explains every result:

| | |
|---|---|
| enum ↔ its exact underlying, both directions | **allowed** (I4 = I4) |
| two *distinct* enums sharing an underlying element type | **allowed** — falls out of the rule, no special case |
| every signedness pair (`int`/`uint`, `byte`/`sbyte`, `long`/`ulong`, `IntPtr`/`UIntPtr`) | rejected |
| `char`/`ushort`, `bool`/`byte`, `float`/`double`, `IntPtr`/`long` | rejected (distinct element types) |
| all widening/narrowing | rejected |
| anything via `Nullable`1` | rejected (exact equivalence, unchanged) |

## Not the array-element rule

Deliberately narrower than `CanCastParam`, already in this file, which *does* collapse signedness. The two genuinely diverge:

- `(uint[])(object)new int[1]` → **OK**;  `(uint)(object)1` → **InvalidCastException**
- `(UIntPtr[])(object)new IntPtr[1]` → **OK**;  `(UIntPtr)(object)(IntPtr)1` → **InvalidCastException**

So `unboxPermitted` is a new predicate, not a reuse of `valueElementNormalisedIdentity`. Both now carry comments pointing at each other so the distinction isn't lost.

## Storage-invariant guard

ECMA-335 II.14.3 permits `bool`-, `char`- and `nint`-backed enums, and the CLR does load them (C# can't declare one; Reflection.Emit can — verified). CoreCLR would permit unboxing those to their underlying, but `CliValueType.IsEnumStructural` covers only the fixed-width integers, so PawPrint stores them wrapped. The interpreter fails loudly in that case.

Relatedly, the two `*HandleInternal` structs *are* in CoreCLR's primitive category as `ELEMENT_TYPE_I` (`methodtablebuilder.cpp`), and PawPrint already flattens them, so they now map to `IntPtr`'s element identity. `RuntimeArgumentHandle` is in that same CoreCLR list but is left unclassified, because PawPrint has no `PrimitiveLikeKind` for it — commented in place.

## Supporting changes

Hoists `isEnumValueType` and `enumUnderlyingHandle` to module level (they were locals of `isConcreteTypeAssignableTo`), mirroring the #672 consolidation. `unboxPermitted` is module-level and named rather than inlined because the `unbox` opcode will need the identical rule when implemented — `Unsafe.Unbox<int>(boxedEnum)` goes through the same helper.

Materialisation now uses the boxed object's handle rather than the target's: under the relaxation the two differ, and the boxed handle is the one `unboxedContents` documents as its precondition.

### DIFF
--- a/WoofWare.PawPrint/IlMachineRuntimeMetadata.fs
+++ b/WoofWare.PawPrint/IlMachineRuntimeMetadata.fs
@@ -1401,6 +1401,50 @@ module IlMachineRuntimeMetadata =
                     | None -> state, None
                     | Some (state, underlying) -> state, builtInPrimitiveIdentity state underlying
 
+    /// Does PawPrint store values of `handle` in the flattened form that the eval stack expects for
+    /// a bare primitive?
+    ///
+    /// True for the built-in primitives themselves, and for enums over the fixed-width integers.
+    /// False for enums over `bool`, `char` or a native int: ECMA-335 II.14.3 permits those and the
+    /// CLR does load them (C# cannot declare one, but Reflection.Emit can), yet
+    /// `CliValueType.IsEnumStructural` deliberately answers false for them, so their storage stays
+    /// a wrapped `CliValueType`.
+    let private unboxMaterialisesFlattened
+        (loggerFactory : ILoggerFactory)
+        (baseClassTypes : BaseClassTypes<DumpedAssembly>)
+        (state : IlMachineState)
+        (handle : ConcreteTypeHandle)
+        : IlMachineState * bool
+        =
+        let state, isEnum = isEnumValueType loggerFactory baseClassTypes state handle
+
+        if not isEnum then
+            // A built-in primitive is flattened by definition; nothing else reaches here, because
+            // the caller has already established a primitive element identity for `handle`.
+            state, true
+        else
+            match enumUnderlyingHandle loggerFactory baseClassTypes state handle with
+            | None -> state, false
+            | Some (state, underlying) ->
+                let state, underlyingId =
+                    primitiveElementIdentity loggerFactory baseClassTypes state underlying
+
+                match underlyingId with
+                | None -> state, false
+                | Some underlyingId ->
+                    [
+                        baseClassTypes.SByte
+                        baseClassTypes.Byte
+                        baseClassTypes.Int16
+                        baseClassTypes.UInt16
+                        baseClassTypes.Int32
+                        baseClassTypes.UInt32
+                        baseClassTypes.Int64
+                        baseClassTypes.UInt64
+                    ]
+                    |> List.exists (fun ty -> ty.Identity = underlyingId)
+                    |> fun flattenable -> state, flattenable
+
     /// CoreCLR `CastHelpers.Unbox_Helper`: `unbox` and the value-typed form of `unbox.any` accept a
     /// boxed operand when the two handles are identical, or when both types are in the primitive
     /// category and report the *same* primitive element type. That second clause is what lets a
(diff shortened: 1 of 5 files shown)


WoofWare.PawPrint/IlMachineRuntimeMetadata.fs
```diff
@@ -1256,6 +1256,286 @@ module IlMachineRuntimeMetadata =
...
+    ///
+    /// The category is CoreCLR's `enum_flag_Category_PrimitiveValueType`, which *includes enums*
+    /// (see the `// Enum is included` remarks in RuntimeHelpers.CoreCLR.cs); an enum reports the
+    /// element type of its underlying integer. Everything else — user structs, `Nullable\`1`,
+    /// `System.Decimal`, reference types, the structural handles — answers `None`.
+    ///
+    /// The identity is returned *exactly*: `Int32` and `UInt32` are different answers, as are
+    /// `Char`/`UInt16`, `Boolean`/`Byte` and `IntPtr`/`Int64`. This is deliberately narrower than
+    /// both ECMA-335's verification types and the array-element rule below, each of which collapses
+    /// signedness — see `unboxPermitted` for why that distinction is load-bearing.
```
Smaug123:
```suggestion
    /// signedness — see `unboxPermitted` for why we make that distinction.
```
I've succumbed to the general Claudiness of the codebase, but there are things up with which I will not put

WoofWare.PawPrint/IlMachineRuntimeMetadata.fs
```diff
@@ -1256,6 +1256,286 @@ module IlMachineRuntimeMetadata =
...
+                    //   - the boxed side drives materialisation, so an unflattened one would push a
+                    //     wrapped `UserDefinedValueType` where the next instruction expects a bare
+                    //     stack primitive;
+                    //   - the target side is the slot the value lands in, and `toCliTypeCoerced`
+                    //     rejects a bare primitive into a value-type slot unless that slot is
+                    //     primitive-like (see the `failwith` in its `CliType.ValueType` arm), so an
+                    //     unflattened target would abort on the following `stloc`/`stfld` instead.
+                    // The identity case never reaches here, so this only ever rejects genuinely
+                    // mixed pairs. Fail loudly rather than answering `false`, which would raise
+                    // InvalidCastException where a real runtime succeeds — a quieter way of being
```
Smaug123:
```suggestion
                    // InvalidCastException where a real runtime would succeed.
```

WoofWare.PawPrint/IlMachineRuntimeMetadata.fs
```diff
@@ -1256,6 +1256,286 @@ module IlMachineRuntimeMetadata =
...
+                    // wrong.
```
Smaug123:
```suggestion
```

WoofWare.PawPrint/IlMachineRuntimeMetadata.fs
```diff
@@ -1256,6 +1256,285 @@ module IlMachineRuntimeMetadata =
...
+        | ConcreteTypeHandle.Pointer _
+        | ConcreteTypeHandle.FunctionPointer _ -> state, None
+        | ConcreteTypeHandle.Concrete _ ->
+            match builtInPrimitiveIdentity state handle with
+            | Some id -> state, Some id
+            | None ->
+                // An enum is in the primitive category too, reporting the element type of its
+                // underlying integer. That underlying must itself be a built-in primitive — the
+                // CLR type loader will not admit an enum whose `value__` is anything else — so
+                // this resolves in exactly one step rather than recursing. Keeping it non-recursive
```
Smaug123:
```suggestion
                // this resolves in exactly one step rather than recursing.
```

WoofWare.PawPrint/IlMachineRuntimeMetadata.fs
```diff
@@ -1256,6 +1256,285 @@ module IlMachineRuntimeMetadata =
...
+                // means malformed metadata answers `None` instead of risking an unbounded walk.
```
Smaug123:
```suggestion
```
You *just* said the semantically correct thing to do was to avoid recursing. Why justify it further?

PR #681, opened 2026-08-04
Implement the Array bound-accessor intrinsics

Smaug123:
Implements `System.Array.GetLength`, `GetLowerBound` and `GetUpperBound`.

## Why not `safeIntrinsics`

All three are `[Intrinsic]`, and every upstream body (`Array.cs` L763-806) bottoms out in `this.GetMultiDimensionalArrayBounds()` — a raw walk over the inline bounds block CoreCLR lays out between the object header and the element data. PawPrint doesn't model that block at all; `RuntimeFieldProjection` fails explicitly on it.

## Semantics

Upstream's `rank` comes from the MethodTable's multi-dim rank field, which is **0** for a szarray — hence its `rank == 0 && dimension == 0` special case. PawPrint stores a szarray as `Lengths = [| totalLength |]`, so indexing `Lengths` uniformly reproduces both the szarray and multi-dim answers with no special case.

Upstream's bound check is the unsigned `(uint)dimension >= (uint)rank`; the signed `dimension < 0 || dimension >= rank` used here is equivalent over the whole of `Int32`, `int.MinValue` included. Every edge case was worked through against the upstream source; no divergent input exists under PawPrint's array model.

## Raising `IndexOutOfRangeException`

`Intrinsics.fs` is compiled before `IlMachineStateExecution.fs`, so the ctor-running `raiseRuntimeException` is unreachable. Options:

1. **The `Enum.HasFlag` precedent in this file** — `allocateRuntimeException` + `overwriteHResultPostCtor` + `throwExceptionObject` (chosen).
2. Give `Intrinsics.call` a return type describing the effect, letting `IlMachineStateExecution` perform it. Rejected: changes a shared signature and every arm in a ~2500-line file, for one exception site.
3. Hoist `raiseRuntimeException`'s transitive dependencies before `Intrinsics.fs`. Rejected as its own refactor — but it's the principled fix if this pattern spreads.

(1) skips the ctor, leaving `_message` null so `.Message` falls back to `"Exception of type X was thrown"`. A new `IlMachineState.setExceptionMessage` — sibling to the existing `setExceptionStackTraceString` — populates it with the string the CLR passes, `SR.IndexOutOfRange_ArrayRankIndex` = *"Array does not have that many dimensions."*, confirmed against the real runtime. The helper is reusable by any future ctor-skipping raise.

### Known gap, commented at the site

(1) also cannot express *unhandled*: `Intrinsics.call` returns `IlMachineState option`, so when `throwExceptionObject` reports `ExceptionUnhandled` the arm can only `failwith`, taking down the interpreter instead of producing `RunOutcome.GuestUnhandledException`. This is **pre-existing** and shared with the `Enum.HasFlag` arms and the three `ExceptionUnhandled` sites in `IlMachineStateExecution`; option (2) is what fixes it generally.

It's deliberately not pinned by a test: `TestPureCases`'s `unimplemented` category asserts the real runtime exits *normally*, and there's no category for "PawPrint can't do this yet **and** the real runtime throws". I tried adding one and removed it rather than distort the harness — flagging in case you'd like that category to exist.


WoofWare.PawPrint/Intrinsics.fs
```diff
@@ -2253,6 +2253,133 @@ module Intrinsics =
...
+            | EvalStackValue.ObjectRef addr ->
+                let arr =
+                    match state.ManagedHeap.Arrays.TryGetValue addr with
+                    | true, arr -> arr
+                    | false, _ -> failwith $"Array.%s{boundKind}: no array allocated at %O{addr}"
+
+                // The stored shape and the receiver's concrete type must agree on rank. Both are
+                // written together by `allocateArray` / `allocateMultiDimArray`, so this cannot
+                // currently fire; it is here to catch representation drift rather than to handle a
+                // reachable case (AGENTS.md: have the machine enforce invariants).
```
Smaug123:
```suggestion
                // reachable case.
```

PR #727, opened 2026-08-04
Implement GC.GetMemoryInfo

### LINKED ISSUE #698: GC.GetMemoryInfo(GCMemoryInfoData, int)
Needs to populate a managed object's fields, not a scalar stub. This is during Async.RunSynchronously thread pool bringup.

Smaug123:
Closes #698.

`GC.GetMemoryInfo(GCMemoryInfoData data, int kind)` is the InternalCall behind the public `GC.GetGCMemoryInfo(GCKind)`. As the issue notes, it has to populate a managed object's fields rather than return a scalar: CoreCLR (`comutilnative.cpp:585`) takes the address of every field of `data` and hands them to `GCHeapUtilities::GetGCHeap()->GetMemoryInfo(...)`, which fills them from `kind`'s `last_gc_info` record.

This adds a new `Native/NativeGc.fs` handler (there was previously no GC native surface at all — only `GC.KeepAlive` as an intrinsic), registered in `NativeDispatch`.

### What it reports, and why

**PawPrint never performs a garbage collection, of any kind, ever.** So for every `kind`, the honest answer to "what did the last GC of this kind record" is "there has never been one" — which is exactly the documented all-zero state that `GCMemoryInfo`'s own doc comment describes, where `Index == 0` is the caller-facing way to detect "no GC of this kind has happened". Unlike a real process asking before its first GC, this is not a transient startup state that later goes stale: it is permanently and unconditionally true here.

This *is* a divergence with the CLR. `_highMemoryLoadThresholdBytes` and `_totalAvailableMemoryBytes` are upstream always derived from the configured heap limit / physical memory, independent of GC history, so reporting them as zero is a combination no real CLR produces. There are two BCL consumers of their ratio:

* `System.Buffers.Utilities.GetMemoryPressure` (used by `ArrayPool<T>`'s trim heuristic), which would read `0 >= 0` as "high pressure" where a real CLR reads "no data yet". The only observable effect is `ArrayPool<T>` trimming more eagerly — not a program-correctness concern, since pool contents were never part of its contract. If a real need appears, the modelling belongs in `EmulatedKernel`, not as an ad hoc constant here.
* System.Threading.PortableThreadPool.OnGen2GCCallback (PortableThreadPool.cs:471-479) reads HighMemoryLoadThresholdBytes and MemoryLoadBytes directly, and the gate thread calls it eagerly at startup (PortableThreadPool.GateThread.cs:58-59) rather than only when a real GC happens. Its result feeds the thread pool's blocking-adjustment heuristic (PortableThreadPool.Blocking.cs:192-226) — the very path that https://github.com/Smaug123/WoofWare.PawPrint/pull/728, stacked directly on top of this, exercises. It is not a crash: Blocking.cs:193 guards with if (memoryLimitBytes <= 0) break;, so our zero cleanly skips the throttle. But the effect is that PawPrint's simulated thread pool never throttles worker-thread growth for memory pressure, silently more permissive than a real CLR under genuine pressure. Harmless today — PawPrint models no memory limit to be under pressure from — but it is the first place to look if simulated and real thread-pool growth ever need to agree, and a reader of the original framing would not have known to look here.

`kind` is still validated: the managed wrapper range-checks it against `[GCKind.Any, GCKind.Background]` before the InternalCall, so an out-of-range value here means a caller bypassed that check, and this asserts rather than silently treating it as one of the four real kinds.

Struct-typed fields (`GCGenerationInfo`, `TimeSpan`) are zeroed by looking up each field's declared signature and taking `CliType.zeroOf` of it, so the zero takes the field's own concrete shape rather than one assumed here that could drift from the real declaration.


Smaug123:
Updated after a further Codex review, which found a real limitation I had missed and I've now pinned rather than papered over.

`GC.GetGCMemoryInfo()` works, but its two **span-valued** properties are unusable past index 0. CoreLib builds them with `MemoryMarshal.CreateReadOnlySpan(ref _generationInfo0, 5)` / `(ref _pauseDuration0, 2)` — a byref to the *first* of a run of sibling fields, walked forward by `sizeof(element)` on the assumption that `[StructLayout(Sequential)]` laid them out contiguously. PawPrint models a heap object as named field cells, not a byte block, so `GenerationInfo[1]` becomes a byte-view read at offset 48 into `_generationInfo0`'s own 32-byte cell:

```
TODO: byte-view read at offset 48 for 8 bytes does not fit in single primitive cell of size 32
  at IlMachineManagedByref.resolveCell (IlMachineManagedByref.fs:1086)
```

(offset 48 = element 1 at 32 bytes + `SizeAfterBytes` at 16.) `Length` and index 0 both work.

I deliberately did **not** work around it here. It is a general "sibling fields are not contiguous" gap that predates this PR — this handler is just the first thing to make an instance of it reachable — and fixing it in the byref machinery would unblock every BCL type using the idiom, rather than one caller. Filed as #729, which also notes the design choice worth making explicitly (step from one field cell into the next sibling, vs. give such objects contiguous backing; the former looks more in keeping with AGENTS.md's "make walks total" guidance).

Added `sourcesPure/GCMemoryInfoSpanProperties.cs` in the `unimplemented` set so it isn't forgotten, and documented the limitation on the handler itself so it's visible in the code and not only in the tracker. The test is differentially safe: it asserts only non-negativity bounds that hold however many real GCs have happened.

Full suite green (1615 passed).

Smaug123:
Correcting a claim in the description above, found by an independent Fable review. The PR body says `ArrayPool<T>`'s trim heuristic is "the one BCL consumer" of `HighMemoryLoadThresholdBytes`/`TotalAvailableMemoryBytes`. **That is wrong**, and the second consumer is more consequential.

`System.Threading.PortableThreadPool.OnGen2GCCallback` (`PortableThreadPool.cs:471-479`) reads `HighMemoryLoadThresholdBytes` and `MemoryLoadBytes` directly, and the gate thread calls it **eagerly at startup** (`PortableThreadPool.GateThread.cs:58-59`) rather than only when a real GC happens. Its result feeds the thread pool's blocking-adjustment heuristic (`PortableThreadPool.Blocking.cs:192-226`) — the very path that #728, stacked directly on top of this, exercises.

It is **not** a crash: `Blocking.cs:193` guards with `if (memoryLimitBytes <= 0) break;`, so our zero cleanly skips the throttle. But the effect is that PawPrint's simulated thread pool **never throttles worker-thread growth for memory pressure**, silently more permissive than a real CLR under genuine pressure. Harmless today — PawPrint models no memory limit to be under pressure *from* — but it is the first place to look if simulated and real thread-pool growth ever need to agree, and a reader of the original framing would not have known to look here.

The handler comment now documents both consumers, and notes that unlike the GC-history fields these two are computed unconditionally from `total_physical_mem` upstream, so reporting them as zero is a genuine divergence rather than the documented never-collected state.

Also addressed a coverage gap the same review flagged: `GCGetMemoryInfo.cs` only touched scalar fields, so a regression in `zeroStructField` — the separate, type-driven recursive-zero path that writes `GCGenerationInfo`/`TimeSpan` fields — would have gone uncaught. It now also asserts `PauseTimePercentage` bounds, both span lengths, and every component of `GenerationInfo[0]` plus `PauseDurations[0]`. Index 0 only, since past that is #729.

Full suite green (1615 passed).

Smaug123:
Rebased onto the new `main` (dropping the six commits that have since merged), and fixed a flake that this PR's own test was carrying.

**The flake.** `sourcesPure/GCGetMemoryInfo.cs` failed intermittently in full-suite runs — twice in about twenty runs, never reproducible in isolation, and not reproducible in 700k iterations of a standalone probe under allocation churn and against a large fragmented heap.

The cause is structural rather than incidental. `GC.GetMemoryInfo` fills the caller's `GCMemoryInfoData` **field by field** out of a `last_recorded_gc_info` record with no synchronisation (`comutilnative.cpp` → `gc.cpp`, `GCHeap::GetMemoryInfo`). A *blocking* GC can't interleave with that, because it suspends the EE — but a **background** GC runs concurrently with managed code, and `GCKind.Background` (plus `GCKind.Any`, when the last recorded collection was a BGC) reads the BGC record through `get_completed_bgc_info()`. A BGC completing mid-call is therefore observable as a torn record: some fields from the old all-zero record, some from the new one.

So every *cross-field* assertion in that file was a race, however obviously true it looks of a coherent snapshot. Three are removed:

- `Index == 0` ⟹ every other field is zero (the documented never-collected state)
- `FragmentedBytes <= HeapSizeBytes`
- `GenerationInfo[0].FragmentationAfterBytes <= SizeAfterBytes`

Each of these genuinely holds of any single coherent record — `gc.cpp:51405-51406` writes `heap_size` and `fragmentation` together at the end of a collection — and each can still be violated by reading two fields either side of a BGC completion. What remains is single-field only: non-negativity, percentage bounds, span lengths.

**Keeping the coverage that mattered.** Stripping those would have gutted the test: PawPrint's actual contract is that *every* field is zero, and a handler returning plausible nonzero garbage would still satisfy non-negativity. But that contract is not a cross-runtime property — a real CLR reports genuine heap sizes — so it cannot be asserted in a `sourcesPure` case at all, since those are diffed against the real runtime's exit code.

It moves to `sourcesImpure/GCMemoryInfoAllZero.cs`, where the expected exit code is PawPrint's alone: all fourteen scalars, both span lengths, all four `GenerationInfo[0]` fields and `PauseDurations[0]`, across all five `GCKind` overloads. It returns 42 rather than 0 on success so "the guest never really ran" can't be mistaken for a pass. The two zeroed fields that are a real divergence from upstream (`HighMemoryLoadThresholdBytes`, `TotalAvailableMemoryBytes` — computed unconditionally from physical memory upstream rather than from any last-GC record) are asserted there explicitly, so introducing a simulated memory budget later has to be a deliberate decision rather than a silent one.

Full suite green at 1628, three consecutive runs.

Issue #726, opened 2026-08-04
SystemNative_GetTimestamp PInvoke (bounds how many blocking thread-pool waits a guest can do)


Smaug123:
Found while adding Task test coverage (#713), on top of the `GC.GetMemoryInfo` work (#698).

```
Unimplemented native method (PInvokeImpl libSystem.Native!SystemNative_GetTimestamp):
System.Private.CoreLib .Sys::GetTimestamp() -> System.Int64
```

`SystemNative_GetTimestamp` is `clock_gettime(CLOCK_MONOTONIC)` — the monotonic tick source behind `Stopwatch.GetTimestamp` / `Environment.TickCount64`. Like the rest of the clock surface it should come from `EmulatedKernel`'s virtual clock, never the host; the existing `SystemNative_GetSystemTimeAsTicks` and `SystemNative_GetLowResolutionTimestamp` arms in `NativeSystemNative.fs` are the obvious precedent, though note this one is *monotonic* rather than wall-clock, so it should be immune to any wall-clock epoch configuration and must never go backwards.

### Why it matters: it puts a hard ceiling on blocking thread-pool waits

Once `GC.GetMemoryInfo` is implemented, a guest can perform several blocking waits on pool-scheduled work — but not many. The thread pool starts consulting the clock once it has adjusted often enough, and then dies here.

Measured on the #698 branch, the boundary is sharp and reproducible:

| Program | Result |
|---|---|
| 3x `Task.Run(...).Result` | passes |
| 4x `Task.Run(...).Result` | **fails** here |
| 5x, 10x `Task.Run(...).Result` | fails here |
| 2x `Task.Run` + awaited `Task.Yield()` | passes |
| 3x `Task.Run` + awaited `Task.Yield()` | **fails** here |
| 2x `Task.Run` + `ContinueWith` | passes |

So the current budget is exactly three blocking pool waits per process. That is enough for focused test cases but will bite any guest doing real concurrent work, so this is probably the next thing to fix in this area.

`sourcesPure/TaskMultipleBlockingWaits.cs` is deliberately sized to three waits and cites this issue; raise it once this lands.

Smaug123:
Fixed in #735

Issue #724, opened 2026-08-04
SystemNative_SchedGetCpu PInvoke


Smaug123:
Found while adding Task test coverage (#713).

`Task.Delay(1).Wait()` reaches the unimplemented PInvoke `SystemNative_SchedGetCpu`; there is no case for it in `NativeSystemNative.fs` (nearest sibling: `SystemNative_GetCpuUtilization`, `NativeSystemNative.fs:195`).

`sched_getcpu(3)` returns the index of the logical CPU the calling thread is currently running on. In a deterministic interpreter that is simulated-process state, not a host read — it should presumably be answered from `EmulatedKernel` and must be consistent with the `ProcessorCount` the guest already observes (i.e. always in `[0, ProcessorCount)`). Whether it should be a constant 0 or track a simulated per-thread affinity is a design question worth deciding explicitly; upstream uses it only to index per-CPU shards, so any value in range is functionally correct, but the choice should be stated.

Parked as `sourcesPure/TaskDelayWait.cs` in `TestPureCases.unimplemented` with this reason.

Issue #723, opened 2026-08-04
ldsflda with a MemberReference token (cross-assembly static field)


Smaug123:
Found while adding Task test coverage (#713).

`UnaryMetadataFieldOps.executeLdsflda` (`WoofWare.PawPrint/UnaryMetadataFieldOps.fs:607`) only handles `MetadataToken.FieldDefinition`. A `ldsflda` whose operand is a `MemberReference` — i.e. the static field is defined in a *different* assembly than the code taking its address — falls through.

Same shape as #691 (`stelem` only handled TypeDefinition/TypeSpecification and needed a TypeReference arm); worth checking whether the sibling `executeLdsfld`/`executeStsfld` have the same gap, and whether there is already a shared token-resolution helper to reuse.

Repro: any `Task.Run` delegate that throws. As soon as it does, Task's internal fault-capture path does `ldsflda` against a cross-assembly static field. It reproduces from a bare `t.IsFaulted` spin-check alone — no `.Wait()` or `.Exception` access needed:

```csharp
Task t = Task.Run(() => throw new InvalidOperationException());
while (!t.IsFaulted) { }
```

Parked as `sourcesPure/TaskRunThrowSetsFaulted.cs` in `TestPureCases.unimplemented` with this reason.

Issue #711, opened 2026-08-04
ldind.ref through a native-int-typed managed pointer


Smaug123:
Split out of #699 as the first concrete blocker on `Task.WhenAny`.

`Task.WhenAny(Task.FromResult(1), Task.FromResult(2))` fails at `NullaryIlOp.fs:2369`:

```
TODO: NativeInt(<managed pointer <variable 2 in method frame <frame #78> of thread 0>>)
```

The `Ldind_ref` arm accepts `EvalStackValue.ManagedPointer src` (reading via `IlMachineState.readManagedByref`) and `EvalStackValue.NativeInt (NativeIntSource.GcHandlePtr _)`, but falls through to `failwith` for `EvalStackValue.NativeInt (NativeIntSource.ManagedPointer src)` — a managed pointer that has been widened to native-int type on the eval stack.

ECMA-335 III.3.44 allows the address operand of `ldind.*` to be `native int` as well as `&`, and the *generic* `ldind` helper in the same file already handles this case (`NullaryIlOp.fs:813`), as does `stind` (`NullaryIlOp.fs:864`). `Ldind_ref` is handled by a separate arm which was simply never taught the native-int form. So this is an inconsistency between `ldind.ref` and its siblings rather than a genuinely missing capability.

Note the primitive `ldind` path routes the native-int case through `readManagedByrefBytesAs`, which is not obviously the right choice for a reference: the pointer here is an ordinary local-variable byref, not a byte-view over stack/native memory, so `readManagedByref` is what's wanted. Whoever picks this up should work out which of the two is correct for each pointer flavour rather than copying one arm onto the other — the existing arm already distinguishes `isStackMemoryPointer` / `isNativeMemoryPointer` / `isTrailingByteViewPointer` from plain byrefs.

Repro:

```csharp
using System; using System.Threading.Tasks;
public static class Program {
    public static int Main(string[] a) {
        Task<int> x = Task.FromResult(1);
        Task<int> y = Task.FromResult(2);
        Task<int> done = Task.WhenAny(x, y).Result;
        return (done == x || done == y) ? 0 : 1;
    }
}
```

Ideally the regression test should be a *minimal* C# case that emits `ldind.ref` on a native-int address directly, with `Task.WhenAny` as a separate end-to-end case, so the test doesn't depend on the whole Task stack.

Smaug123:
[...]

PR #728, opened 2026-08-04
Add Task end-to-end test coverage

### LINKED ISSUE #713: No end-to-end test coverage for Task at all
Noticed while decomposing #699. There is not a single `Task` test case in `sourcesPure`/`sourcesImpure` — the only concurrency coverage is `Thread*`.

That matters because rather more works than I expected. Verified by hand against `6e43cf4`, all exiting 0 under PawPrint:

- `Task.FromResult(41).Result`
- `TaskCompletionSource<int>` — `SetResult` then `.Task.Result`
- **`Task.Run(() => 41).Result`** — i.e. thread-pool dispatch works
- **`async Task<int>` with `await`, then `.Result`** — i.e. the state-machine/awaiter path works

None of that is pinned, so any of it could regress silently. Worth a `sourcesPure` case (or a few) locking in the current baseline before more Task work lands on top of it.

Known-failing, tracked separately: `Task.WhenAny` (#711), `Task.WhenAll` (#712).

Smaug123:
Closes #713.

Rebased onto current `main`; the stack beneath it (#711, #725, #727) has landed, so this is now a single-purpose branch. **Test-only**: no production code changes.

There was not a single `Task` case in `sourcesPure`/`sourcesImpure` — the only concurrency coverage was `Thread*` — despite rather a lot already working. Everything pinned here was verified to pass *before* this commit; nothing in the interpreter changed to make it pass.

### Pinned as working

`Task.FromResult(x).Result` · `TaskCompletionSource<int>` · `Task.Run(() => n).Result` (thread-pool dispatch) · `async Task<int>` + `await` + `.Result` · `await Task.Yield()` · `Task<T>.ContinueWith` · `async void` (synchronised via `ManualResetEventSlim`, no timing assumptions).

One behaviour per file rather than one bundled case, so a failure names the behaviour that regressed rather than just "TaskBasics".

### The multiple-blocking-wait path, and why the count is still 3

`TaskMultipleBlockingWaits.cs` pins what `GC.GetMemoryInfo` (#727) unblocked: a *second* blocking wait on pool-scheduled work in the same process reaches the thread pool's blocking-adjustment heuristic, which calls that InternalCall. Before it was implemented, every Task case could afford only one such wait — which is why the sibling files are each written to perform exactly one.

That budget is still bounded above, so this case is sized to the measured frontier rather than a round number. **Re-measured after the rebase**: three blocking pool waits pass, four fail — the boundary is unchanged, but the primitive it stops at has moved. It used to be `SystemNative_GetTimestamp` (#726); since that landed in #735, the fourth wait instead gets as far as the pool's hill-climbing controller adjusting its thread count, which calls `Math.Pow` (`PortableThreadPool.HillClimbing.cs:301`) — an unimplemented JIT intrinsic.

| Program | Result |
|---|---|
| 3x `Task.Run(...).Result` | passes |
| 4x, 6x, 8x, 10x `Task.Run(...).Result` | fails (`Math.Pow`) |
| 2x `Task.Run` + awaited `Task.Yield()` | passes |
| 3x `Task.Run` + awaited `Task.Yield()` | fails (`Math.Pow`) |

Filed as #755; the test comment now says to raise the count once it lands.

### Found while probing, parked rather than quietly omitted

Both cases were originally parked on blockers that have since landed, so both were re-run against current `main`. Both still fail, but at a *new* primitive; the second commit updates the stated reasons rather than leaving them to mislead the next reader.

- `TaskRunThrowSetsFaulted.cs` — the `ldsflda`-via-`MemberReference` gap (#723) landed in #740, so Task's fault-capture path now reaches the unimplemented `ExceptionNative_GetFrozenStackTrace` QCall, filed as #754. Re-verified that a bare `t.IsFaulted` spin-check still reproduces it, so this is still inside Task's own fault-recording.
- `TaskDelayWait.cs` — `SystemNative_SchedGetCpu` (#724) landed in #742, so `Task.Delay(1).Wait()` now reaches `TimerQueueTimer..ctor`, which indexes `TimerQueue.Instances` by `(long) Thread.GetCurrentProcessorId() % Instances.Length` and narrows with `conv.ovf.i` before the `ldelem.ref`. `Conv_ovf_i` is unimplemented (`NullaryIlOp.fs:2368`), filed as #753; the adjacent `Conv_ovf_u` is implemented and shows the shape of the fix.

Both are in `unimplemented` with precise reasons citing file:line, and all three new blockers are now filed: #753 (`conv.ovf.i`), #754 (`ExceptionNative_GetFrozenStackTrace`), #755 (`Math.Pow`).

### Determinism

Every assertion is on a returned value: never on which worker thread ran something, nor on timing, nor on ordering between independent tasks. These are differential tests run against the real runtime as well as PawPrint, so an assertion holding only under the simulated scheduler would be a latent flake.

Full suite green (1731 passed).


Smaug123:
Note for when this lands: #736 implements `SystemNative_GetTimestamp`, which removes the ceiling `sourcesPure/TaskMultipleBlockingWaits.cs` is sized against. Per #726's closing instruction the wait count here should be raised and the #726 citation dropped once both are in. Left to this PR rather than done in #736, since the file exists only on this branch.

PR #735, opened 2026-08-05
Implement SystemNative_GetTimestamp, unblocking Stopwatch

Smaug123:
First of a two-PR stack for #724. This one lands the dependency; the follow-up (per-thread simulated CPU + `SystemNative_SchedGetCpu`) stacks on top.

## Why this is a prerequisite for #724

Implementing `SystemNative_SchedGetCpu` with a *valid* (non-negative) CPU id makes CoreLib run `ProcessorIdCache.ProcessorNumberSpeedCheck()` during static initialisation, which times `Thread.GetCurrentProcessorNumber()` against a ThreadStatic read using `Stopwatch.GetTimestamp()`. That reaches `SystemNative_GetTimestamp`, which was unimplemented — so #724 could not be landed with a real value until this exists. (Returning `-1` from `SchedGetCpu` dodges the speed check entirely, but that is the "sched_getcpu unsupported" path and defeats the point of tracking per-thread CPU state.)

It is worth landing separately regardless: it unblocks all of `Stopwatch` for every guest.

## What it does

`Stopwatch.GetTimestamp()` on a Unix CoreLib is `Interop.Sys.GetTimestamp` -> `SystemNative_GetTimestamp`, which upstream is a one-line forward to `minipal_hires_ticks()` — `clock_gettime_nsec_np(CLOCK_UPTIME_RAW)` on macOS, `clock_gettime(CLOCK_MONOTONIC)` scaled to nanoseconds on Linux. (`pal_time.c` and `minipal/time.c` are both outside our sparse `dotnet/runtime` checkout, so they were read by fetching them at the pinned commit.)

The reading derives from the same `EmulatedKernel.VirtualClockMs` that already backs `SystemNative_GetLowResolutionTimestamp`, via a new `EmulatedKernel.monotonicTimestampNanos`. That is not merely convenient: upstream, the low-resolution entry point is `minipal_lowres_ticks()`, which reads *the very same monotonic clock* in milliseconds. Sharing one field reproduces a relationship the guest can observe — `Environment.TickCount64` and `Stopwatch` must not disagree about how much time has passed.

The units are not ours to choose: `minipal_hires_tick_frequency()` is the constant 1e9, and `Stopwatch.GetFrequency()` hard-codes that same constant on Unix.

### Overflow guard

`maxMonotonicTimestampClockMs` (~292 years of simulated uptime) is deliberately *tighter* than the wall clock's `maxWallClockEpochMs`: a virtual-clock reading that `systemTimeAsTicks` still accepts can be too large for the nanosecond derivation. Silently wrapping into a negative timestamp would make every `Stopwatch` in the guest report nonsense, so the derivation checks rather than assumes.

## Consequences worth stating

- **Millisecond granularity.** The virtual clock advances 1 ms per scheduler tick, so every timestamp the guest observes is a multiple of 1,000,000 ns. `Stopwatch` is not a source of unique values here — but real `clock_gettime(CLOCK_MONOTONIC)` makes no uniqueness guarantee either, so this is a faithful gap rather than one to paper over.
- **Weak monotonicity.** Reading never advances the clock, so two reads within one scheduler tick are equal and a zero-length measured interval is normal. Same property `VirtualClockMs` already guarantees for `Environment.TickCount64`.
- **Boots at zero**, where real CLOCK_MONOTONIC counts from an unspecified origin. That is part of the replay contract.


Smaug123:
Updated after comparing against #736, which implements the same entry point independently. The two agreed on essentially everything — same derivation from `VirtualClockMs`, same `nanosecondsPerMillisecond`, same overflow bound `9_223_372_036_854L`, same fail-loud-rather-than-saturate guard at the point of use — but #736 was **right about something this PR had wrong**, and had two properties this one was missing. Ported:

**Corrected: the overflow horizon is reachable by ordinary guest code.** This PR previously claimed it was "unreachable in practice". It isn't. A sleep deadline is `VirtualClockMs + timeout` with no cap, and when no thread is Runnable the driver's deadline jump moves the clock the whole way there. Measured under PawPrint: eight `Thread.Sleep(int.MaxValue)` calls advance the clock by 17,179,869,451 ms, so roughly 4,300 of them cross the bound. The doc comment now says so, and notes the related asymmetry — the bound is ~27× tighter than `maxWallClockEpochMs`, so there is a band of readings from which `DateTime.UtcNow` is derivable but `Stopwatch.GetTimestamp` is not, because each clock-derived PAL entry enforces its own ceiling lazily rather than the scheduler bounding `VirtualClockMs` centrally. That central fix is worth doing and is deliberately out of scope here, since it would change the failure point of existing behaviour (`systemTimeAsTicks` has the same lazy shape today).

**Added two properties** this PR could not have caught, because its kernel helper always left `WallClockEpochMs` at zero:
- *the wall-clock epoch cannot perturb the monotonic clock* — `CLOCK_MONOTONIC` is immune to wall-clock changes, and here that is structural rather than enforced; worth pinning because a future NTP-skew model would be at risk of quietly breaking it.
- *the monotonic and wall clocks agree about elapsed time* — a guest that times an interval with `Stopwatch` and one that times it with `DateTime.UtcNow` must get the same answer.

**Sharpened** the overflow-bound test to exhibit the actual int64 wrap (`shouldBeSmallerThan 0L`) rather than sidestepping it via `decimal`, since a negative product is precisely the failure the bound exists to prevent.

**Also noted** #736's additional motivation in the handler comment: the thread pool's hill-climbing step times its own sampling window with this entry point (`PortableThreadPool.AdjustMaxWorkersActive`, PortableThreadPool.cs:379), not just `ProcessorIdCache`.

Retained from this PR and not present in #736: the impure test pinning the PawPrint-specific replay contract (boots at zero, whole-millisecond granularity, `TickCount64` and `Stopwatch` are one clock) — a pure case structurally cannot pin those, being diffed against the real runtime; the property tying `GetTimestamp` to `GetLowResolutionTimestamp`, which upstream are `minipal_hires_ticks` and `minipal_lowres_ticks` reading the same clock; and the handler comment's full upstream chain including the macOS `clock_gettime_nsec_np(CLOCK_UPTIME_RAW)` branch.

Full suite green (1663 passed).

PR #768, opened 2026-08-06
Implement the OS thread id PAL entry points

Smaug123:
Implements `SystemNative_TryGetUInt32OSThreadId` and `SystemNative_GetUInt64OSThreadId`.

## Why both

CoreLib's `Lock.ThreadId.InitializeForCurrentThread` (`Lock.NonNativeAot.cs`) is `#if`-split per target, and each flavour declares only its own `Interop.Sys` stub — confirmed by dumping both:

| CoreLib | declares |
|---|---|
| linux-x64 | `TryGetUInt32OSThreadId() : uint32` only |
| osx-arm64 | `GetUInt64OSThreadId() : uint64` only |

Upstream both read the same `minipal_get_current_thread_id()`; the 64-bit entry returns it verbatim and the 32-bit one truncates (substituting `(uint32)-1` if the truncation is 0). So these are two width-projections of one value, not two features, and implementing one would leave `System.Threading.Lock` dead on half our machines. `TestPureCases.fs`'s note on `TaskDelayWait.cs` already recorded that a fix must cover both.

## The value is an identity

`Lock` uses this for mutual exclusion *and* recursive re-entry detection, so two live threads sharing an id is a silent correctness failure rather than a crash. Two sentinels must also be dodged: `(uint32)-1` means "platform cannot determine a thread id", and CoreLib maps a `0` id to `0xFFFF_FFFF` by decrement — so either would become a *shared* id.

Ids are minted at thread creation and stored in `ThreadState.OsThreadId`. There is one producer, `EmulatedKernel.osThreadId`, a function of the thread's `ThreadId`:

```fsharp
let osThreadId (thread : ThreadId) : OsThreadId = ... OsThreadId (uint32 i + 1u)
```

Uniqueness across live threads — the only property anything needs — is inherited from `ThreadId`, which is already unique and never reused. Every thread PawPrint creates has one, guest-visible and interpreter-internal alike, so there is no second namespace to stay disjoint from and no encoding to defend. The signal dispatcher needs no special case: it runs *guest* handler code (`SignalDispatch` wakes it onto a managed handler), so an aliased id would let a handler take a `Lock` the guest thread already holds — but it gets a distinct `ThreadId` like anything else, so it gets a distinct id for free.

No new cursor is needed either: all three allocation sites already advance `NextThreadId` by exactly one, so a parallel ordinal would be provably equal to it, and redundant state can only drift.

Sentinels: the `+ 1` dodges `0`, which is immediate and real (thread id `0` is the entry thread). The `(uint32)-1` sentinel is unreachable by construction rather than by a bound check — the largest id an `int` `ThreadId` can produce is `0x8000_0000`. A negative `ThreadId` is rejected, since `-1` would wrap onto exactly that fatal `0`; `FrameId -1` is an established sentinel in this codebase, so it is a plausible mistake.

## Why this deliberately differs from CPU placement

`EmulatedKernel.cpuForRotation` must *not* key off `ThreadId` — that is why `NextCpuRotation` exists. `osThreadId` makes the opposite choice, and the difference is what the guest can do with the number:

- A `CpuId` is drawn from a small cyclic range and compared *between* threads. Two threads sharing a core is a meaningful, observable fact, so letting an interpreter-internal allocation shift the rotation changes guest-observable behaviour.
- A thread id is opaque. No BCL code does anything with it but test it for equality, so *which* number a thread gets is not observable — only whether two threads share one.

So the dispatcher, minted lazily on the guest's first `SystemNative_InitializeTerminalAndSignalHandling`, does consume a thread id and does shift the ids of guest threads created after it. Real Linux behaves identically: its signal-handling thread is an ordinary `pthread_create` and consumes a tid like any other. `TestOsThreadId`'s `a parked thread does consume an id, unlike a rotation slot` pins the contrast against `TestCpuPlacement`'s mirror-image test, so the asymmetry reads as a decision rather than an oversight.

`NextCpuRotation` accordingly keeps its name and its CPU-only meaning.


Smaug123:
Blocked on #777 — please merge that first.

Codex review of this branch (after the rebase) found that `Lock.ThreadId` caches the OS thread id in `[ThreadStatic] t_threadId`, and PawPrint stored all statics in one process-wide map. I verified it: on `main` a worker checking `IsHeldByCurrentThread` for a lock the main thread holds fails *loudly* at the unimplemented `SystemNative_GetUInt64OSThreadId`; with this branch's entry points but without per-thread statics, it silently returns `true`.

So merging this first would convert a loud failure into a silent mutual-exclusion bug. #777 gives `[ThreadStatic]` fields real per-thread storage; once it lands I'll rebase this on top and re-verify.

Smaug123:
Rebased onto `main` now that #777 has landed (`6691e2e`), and re-verified as promised: the cross-thread `Lock` scenario that was silently wrong now answers correctly, and `sourcesPure/LockHeldByOtherThread.cs` pins it against both failure modes. Details in the PR body. Full suite 1844/0; Codex found nothing.

Unblocked from my side.

PR #786, opened 2026-08-07
Bundle the host's launch inputs into a HostConfig record

Smaug123:
No-op refactor, split out ahead of the AppContext-from-runtimeconfig work.

`Program.run` and `prepare` took the runtime directories, kernel config, PCT seed
and argv as four positional parameters — already the point at which a caller can
silently transpose two of them, and the AppContext work wants to add a fifth.
They now take a single `HostConfig`.

### Why `KernelConfig` is contained rather than absorbed

`KernelConfig` is the *simulated process's* kernel-visible state: what the guest
would learn by asking the operating system, held as data precisely so a replay
never depends on the machine that recorded it. `HostConfig` is one level out — how
the host launches that process at all — and contains `KernelConfig` as the part of
the launch that describes the kernel. `Argv` sits in `HostConfig` rather than
`KernelConfig` for the same reason: the runtime hands it to `Main` directly, rather
than the guest reading it back through a syscall.

The program image stays a separate parameter throughout, since it's the payload
rather than configuration.

`HostConfig.Default dirs` covers the common case, so most call sites get shorter
rather than longer; the handful that vary something name only what they vary.

### On it being a no-op

`prepare` destructures the record into the same four bindings its body already
used, so the body is untouched — the library diff is 16 lines and provably
behaviour-preserving by inspection.

The call-site rewrite was scripted, so I checked it mechanically rather than
trusting the tests: a transposition of `Kernel`/`PctSeed`/`Argv` would *not*
necessarily fail anything, because most sites pass the defaults. A verifier
recovered the four config arguments from both sides of the diff and required them
to match at all 21 rewritten sites, and the count of `Program.run`/`prepare`
mentions is 28 before and after.

(That was worth doing: the first attempt at the rewrite split `(Some sourceName)`
into two tokens and shifted every field, which built fine in the test project.)

### Documentation

Codex review caught that the scripted rewrite had also matched inside two prose
comments and mangled them; both are restored. It also flagged
`docs/user/fuzz-over-thread-scheduler.md`, which told readers to pass the seed
positionally. Fixing that surfaced that `docs/user/run-a-program.md` was already
stale on two counts unrelated to this PR — it still showed `nativeImpls` and a bare
`environmentVariables` parameter, and its exit-code extraction predates the int32
stack slot carrying provenance — so the example now compiles again.

1964 tests pass.


WoofWare.PawPrint/HostConfig.fs
```diff
@@ -0,0 +1,50 @@
...
+/// Every field here is part of a run's replay contract. Two runs with equal
+/// `HostConfig` over the same image must produce the same trace, so nothing in
+/// here may be filled in from an ambient host read.
+type HostConfig =
+    {
+        /// Directories searched, in order, for framework assemblies. Binding is
+        /// by simple name and takes the first hit, so putting a runtime pack at
+        /// the head of this list selects that pack's CoreLib flavour — see
+        /// `TestLinuxCoreLibFlavour` and the CoreLib-flavour section of
+        /// AGENTS.md.
```
Smaug123: ugh this is a public facing docstring

PR #794, opened 2026-08-07
Seed AppContext from runtimeconfig.json

Smaug123:
BCL feature switches — `EventSource.IsSupported`, `Globalization.Invariant` and friends — are read through `AppContext.GetData`/`TryGetSwitch`, but nothing in managed code populates that store. On CoreCLR the VM calls `AppContext.Setup(char**, char**, int)` from `CorHost2::CreateAppDomainWithManager`, passing arrays `hostpolicy` built from `runtimeOptions.configProperties`. PawPrint never did, so every guest saw an empty store and every switch took its default.

`AppContextSeed.fs` synthesises the two `char**` buffers and calls `Setup`, so CoreLib's own IL runs: the `Dictionary`, the pointer walk and the `new string(char*)` per entry are all its code. The host's only contribution is the buffers, which is exactly the division of labour on a real runtime. (`new string(char*)` running real IL is what #783 landed.

## What this deliberately does not do

Seeding covers `configProperties` and nothing else. A real host populates eight further properties itself — `TRUSTED_PLATFORM_ASSEMBLIES`, `APP_CONTEXT_BASE_DIRECTORY` and friends — from deps resolution and its filesystem layout, before it reads `configProperties` at all; PawPrint has neither, so it populates none of them. That is observable (`AppContext.GetData "TRUSTED_PLATFORM_ASSEMBLIES"` is null here and a path list on CoreCLR) and now recorded in `docs/divergences.md`, along with the related fact that a missing `runtimeconfig.json` is *not* the no-properties case on a real host: hostfxr treats it as a self-contained app and dies before managed code (exit 131).

`EventSourceDisabled.cs` — the motivating case — is **parked**. Seeding works and the guest reaches `Main` with the switch off; what stops it is EventSource's manifest handling comparing two byrefs into the same string at different offsets, which `EvalStackValueComparisons.cltUn` refuses even though they plainly share a root. That is its own change.


WoofWare.PawPrint/RuntimeConfig.fs
```diff
@@ -0,0 +1,303 @@
...
+///
+/// On a real runtime these come from the app's `runtimeconfig.json`: `hostpolicy` reads
+/// `runtimeOptions.configProperties` and hands the result to `AppContext.Setup` as a pair of
+/// `char**` arrays. By the time managed code can see them every value is a string, whatever
+/// its JSON type was — `AppContext.GetData` returns `object`, but the only thing ever stored
+/// is a string. That is why this is a `Map<string, string>` and not a map to some JSON value
+/// type: the string *is* the property, and the JSON type only decides which string.
+///
+/// Construct one with `RuntimeConfig.parse` (which applies hostpolicy's mapping) or with
+/// `AppContextProperties.ofMap` (for a host that knows its own mind, chiefly tests). The
```
Smaug123, line 17: how poetic

Issue #802, opened 2026-08-08
DereferenceFieldById byte-renders overlapping fields, so an overlapped pointer field cannot be read


Smaug123:
`CliValueType.DereferenceFieldById` resolves a field that overlaps others by byte-rendering the storage:

```fsharp
match affectedFields with
| [] -> failwith "unexpectedly didn't dereference a field"
| [ f ] -> f.Contents
| _ :: _ :: _ ->
    let fieldBytes = CliValueType.BytesAt targetField.Offset targetField.Size cvt
    CliType.OfBytesLike targetField.Contents fieldBytes
```

`BytesAt` cannot render a cell carrying pointer provenance, so reading an overlapped `IntPtr`/`UIntPtr` field throws from the host rather than returning the pointer. The single-covering-field case (`[ f ] -> f.Contents`) is unaffected, which is why this only shows up under `[StructLayout(LayoutKind.Explicit)]` with genuinely overlapping fields.

Repro shape:

```csharp
[StructLayout(LayoutKind.Explicit)]
struct Union
{
    [FieldOffset(0)] public long Raw;
    [FieldOffset(0)] public IntPtr Handle;
}

var u = new Union();
u.Handle = Marshal.AllocHGlobal(16);   // a managed pointer with provenance
var read = u.Handle;                    // <-- byte-renders the overlapping storage
```

This is not marshalling-specific — any read of an overlapped pointer field goes through the same path — but it is reachable from `Marshal.StructureToPtr` on such a struct, which is where it was found (`StructMarshalStub.tryComputePlan` resolves each field's effective value through `DereferenceFieldById`, precisely so overlapping fields marshal the winning write; see `sourcesPure/MarshalStructureToPtrDateTimeOverlapping.cs` for the non-pointer case that works).

Note the asymmetry worth preserving in any fix: the two-or-more-overlapping case exists to answer "which write actually covers these bytes", and for byte-addressable values the byte rendering *is* the answer. What is missing is the case where the winning write is a whole cell that exactly covers the requested range — there the cell itself should be returned, as the single-field arm already does, without any byte rendering.

Related: #801 (the same "provenance-carrying cells versus byte operations" tension, in native memory rather than in value-type storage).

Smaug123:
A second consumer, and a note for whoever fixes this: the struct-marshal stub (#819) hits this
path, and there is a *second* obstacle waiting immediately behind it.

`StructMarshalStub.tryComputePlan` resolves each field's effective value through
[...]

Issue #801, opened 2026-08-08
Pointer cell written by the struct-marshal stub cannot be read back, though Marshal.WriteIntPtr's can


Smaug123:
Found while implementing the struct-marshal stub (`Marshal.StructureToPtr` for a has-layout-non-blittable struct).

A pointer-valued cell that the struct-marshal stub writes into `AllocHGlobal`'d memory cannot be read back by the guest, even though the identical value at the identical offset in the identical block *can* be read back when the guest wrote it itself.

## The discriminator

This passes:

```csharp
IntPtr scratch = Marshal.AllocHGlobal(16);
IntPtr p = Marshal.AllocHGlobal(32);
Marshal.WriteIntPtr(p, 8, scratch);
if (Marshal.ReadIntPtr(p, 8) != scratch) return 2;   // reads back fine
```

This fails:

```csharp
[StructLayout(LayoutKind.Sequential)]
struct WithPointerAndDate { public int Id; public IntPtr Handle; public DateTime When; }

var s = new WithPointerAndDate { Id = 7, Handle = scratch, When = new DateTime(2020, 1, 2) };
IntPtr p = Marshal.AllocHGlobal(Marshal.SizeOf<WithPointerAndDate>());
Marshal.StructureToPtr(s, p, false);
Marshal.ReadIntPtr(p, 8);      // <-- dies
```

```
System.Exception : MemoryBlock.readBytes: refusing byte view over native int with
non-byte-addressable provenance <managed pointer <byte 0 of <native memory block #0>>>
at offset 8 in <native memory block #1>
   at WoofWare.PawPrint.MemoryBlockModule.readBytes (MethodState.fs:349)
   at WoofWare.PawPrint.NativeMemoryPoolModule.readBytes (MethodState.fs:579)
   at WoofWare.PawPrint.IlMachineManagedByref.readNativeMemoryBytesAs (IlMachineManagedByref.fs:877)
   at WoofWare.PawPrint.IlMachineManagedByref.readManagedByrefBytesAs (IlMachineManagedByref.fs:1070)
   at WoofWare.PawPrint.NullaryIlOpModule.executeLdind (NullaryIlOp.fs:1104)
```

Same value, same provenance, same block, same offset, same read. So the difference is in *how the cell was installed*, not in the read: the reader byte-reads and trips over a cell in the way, where for the guest-written buffer it resolves the cell instead.

## What has been ruled out

Three plausible explanations, each tested and eliminated:

[...]

PR #803, opened 2026-08-08
Marshal a DateTime field as an OADate through a struct-marshal stub

Smaug123:
Un-parks `MarshalStructureToPtrDateTimeField.cs` by implementing the arm of
`MarshalNative_TryGetStructMarshalStub` that PawPrint refused outright: a struct that has
layout but is not blittable, which today means one holding a `System.DateTime`.

CoreCLR's arm 2 (marshalnative.cpp:118) synthesises an IL stub and returns its entry address;
CoreLib invokes it via `calli` through
`delegate*<ref byte, byte*, int, ref CleanupWorkListElement?, void>`. A `DateTime` field is
diverted to `MARSHAL_TYPE_DATE` (mlinfo.cpp:1747) and becomes an 8-byte OADate double, not the
managed `_dateData` bytes — which is what the parked test asserts, including a belt-and-braces
check that the bytes are *not* the tick count.

## Design

PawPrint has neither IL synthesis nor code addresses, so the stub pointer is a new
`NativeIntSource.StructMarshalStub` carrying the type's identity. Structural equality gives it
the per-type identity CoreCLR's stub cache has, and it answers `isZero` with `false` — load-
bearing, because the guest branches on `structMarshalStub != null` and would otherwise memmove
the managed bytes silently. `calli` on one runs `StructMarshalStub.executeStubCall` rather than
pushing a frame.

The alternative considered and rejected was fabricating a `MethodInfo`. That needs a
`MethodDefinitionHandle`, which is used as an *identity* across `MethodHandleRegistry`,
`IntrinsicMethodKeys` and `MethodInfo.NominallyEqual`; there is no precedent in the codebase for
inventing one, and a lie in an identity field is much harder to reverse than an extra `calli`
arm. Also rejected: intercepting `Marshal.StructureToPtr` itself, which AGENTS.md rules out —
implement the primitive, not the managed method that calls it.

Offsets come from `CliValueType.TryComputeMarshalLayout`, a refactor of the walk that already
answered `Marshal.SizeOf` and then discarded its per-field offsets. Sharing it is what stops a
field landing in one place for sizing and another for writing.

### Calling the guest

The stub is not self-contained: `ToOADate` has a zero special case, a VB compatibility fixup and
a guest-visible `OverflowException`, all of which belong to the guest's CoreLib. So it calls the
guest's own `StubHelpers.DateMarshaler.ConvertToNative`, as CoreCLR's date marshaller does.

`NativeHandlerResult.PushedManagedCallee` cannot serve: that machinery is bound to
`MethodInfo`-backed frames — `executeOneStep` only reaches `dispatchNative` for
`InternalCall`/`PInvoke`, and re-entry re-locates the handler from the live frame's
`ExecutingMethod` — and a stub has no `MethodInfo` by construction. Instead the whole `calli`
re-executes, once per conversion: each pass pushes a callee and leaves the program counter
unadvanced, and the result accumulates on the eval stack above the stub pointer, so the run of
`Float`s above it counts how far we got. Nothing is written until every conversion completes.

## Scope

`CopyBytes` steps are deliberately narrow. A composite field is accepted only where its interior
is trivial (a primitive-like wrapper), because writing the managed value verbatim otherwise
assumes the managed and marshal layout walks agree on that interior, which nothing enforces —
and this arm exists precisely because CoreCLR repositions some fields between the two forms.
Fields carrying a `[MarshalAs]` descriptor are refused outright: width cannot separate the legal
pairings from the illegal ones (`Int32` admits only I4/U4, `DateTime` only DEFAULT/STRUCT), so
classifying on width would silently marshal types the real runtime refuses to load.

Unchanged and still refused loudly: Decimal fields, nested composites needing a recursive plan,
`Unmarshal` (nothing reaches it — `MarshalPtrToStructure.cs` is parked behind an unrelated
`Activator` blocker), and blittable sequential *classes*, which `main` refuses identically.

## Tests

Property tests over the shared layout walk, including an explicit-offset sweep whose offsets are
shuffled out of declaration order — the arm a sequential-only sweep never reaches. Three
end-to-end differential cases covering the single conversion, the multi-conversion retry loop,
`Cleanup`, padding, buffer reuse, a live pointer field, and overlapping explicit layout.

Every claimed behaviour was mutation-tested: breaking the OADate write, reversing conversion
order, dropping the image zeroing, restoring the no-op cleanup, marshalling stale overlapping
contents, and drifting the layout cursor from its placement each fail the intended assertion.

## Review

Four passes (workflow `code-review` twice, `codex review` twice): 18 findings, 14 fixed,
2 argued, 2 filed. The substantive fixes, each with a test that fails without it:

- Read the source struct structurally rather than through its `ref byte` view. A byte read
  flattens every cell, and an `IntPtr` from `Marshal.AllocHGlobal` has no byte rendering, so an
  ordinary struct crashed several BCL frames below the QCall that had just promised it was
  marshalable.
- Zero the native image. CoreCLR opens its Marshal stream with `initblk` over the whole native
  size (dllimport.cpp:1290) and closes its Cleanup stream with the same (:1319); we wrote only
  field ranges and made Cleanup a no-op, so padding kept stale bytes and `DestroyStructure` left
  the old image intact. The real runtime confirms both.
- Marshal each field's *effective* value. Overlapping `[FieldOffset]` fields leave their siblings'
  stored contents stale by design; the real runtime writes the later-declared field's bits, and
  we wrote the OADate then a stale zero over it.
- Write the image field by field, so a reused buffer's pointer cell is replaced rather than
  byte-walked, and write a bare pointer rather than the `IntPtr` wrapper struct.
- `loadClass` before the guest helper call; assert the callee left the program counter
  unadvanced; apply the EXPLICITTHIS correction in the call-site arity check.

Filed rather than fixed, both needing shared machinery that does not exist and both failing
loudly: #801 (a marshalled pointer cell cannot be read back; buffer reuse across differing
layouts) and #802 (`DereferenceFieldById` byte-renders overlapping fields, so an overlapped
pointer field cannot be read — pre-existing and not marshalling-specific).


Smaug123:
## Design note: the driver is provisional

Flagging this so review attention goes to the right places. The `calli` re-execution driver in
this PR — `tryRecognise` plus the resumption half of `executeStubCall`, roughly 440 of the 780
lines in `StructMarshalStub.fs` — is a known local optimum, and the plan is to replace it.

**The weakness.** The resumption record is *inferred* from the evaluation stack rather than
stored: `tryRecognise` recovers "how many conversions have completed" by counting the run of
`Float`s above the stub pointer. That idiom is borrowed from `NativeRuntimeTypeQCall`, where it
is sound because a native frame has its own empty evaluation stack, so a marker on it is
unambiguous by construction. Here the stack belongs to CoreLib's `StructureToPtr` frame, and the
marker is unambiguous only because I checked the surrounding values. The review thread argued
the misroute hazard away on reachability grounds, which is a claim about what else exists today
rather than about the design.

**When it breaks.** Sooner than "eventually". Of CoreLib's field marshallers, `DateMarshaler` is
the only one returning a `Float`:

```
DateMarshaler::ConvertToNative(DateTime)           : double
AnsiCharMarshaler::ConvertToNative(char,bool,bool) : uint8
CSTRMarshaler::ConvertToNative(int32,string,intptr): intptr
CSTRMarshaler::ConvertFixedToNative(...)           : void
FixedWSTRMarshaler::ConvertToNative(...)           : void
```

Widening the scan to `NativeInt` would be fragile but survivable. The **void-returning** ones are
a hard break: a conversion that leaves nothing on the stack cannot be counted by a mechanism
that counts stack values. `HandleMarshaler.ConvertSafeHandleToNative` additionally consumes the
`ref CleanupWorkListElement` argument this PR reads and ignores. So the shape "one call, one
result, accumulate on the stack, nothing owns resources" is specific to `DateTime`.

**The intended replacement.** Widen `MethodInfo`'s identity from a bare `MethodDefinitionHandle`
to admit synthesised methods, add `RuntimeBehaviour.StructMarshalStub`, and dispatch the stub in
`AbstractMachine` alongside `DelegateCtor`/`DelegateInvoke`. It then has its own frame, its own
empty evaluation stack and its own locals, so re-entry uses the existing native-frame machinery
legitimately and `calli` needs no fork at all. That identity widening is a standalone change with
other consumers already waiting — `MarshalPtrToStructure.cs` is parked on exactly it (CoreCLR's
allocator is a JIT helper with no managed `MethodInfo`), as are delegate `BeginInvoke`/`EndInvoke`
and the multi-dim array helpers — so it will land as its own PR, with the port of this stub as a
small follow-up.

**What is not provisional**, and is worth reviewing properly, is everything the replacement
builds on: `CliValueType.TryComputeMarshalLayout` and the sharing of one layout walk between
sizing and writing; `StructMarshalPlan` / `tryComputePlan` and the classifier they share with the
blittable arm; `NativeIntSource.StructMarshalStub` and its `isZero` contract; and the
marshalling semantics themselves (OADate conversion via the guest's own helper, whole-image
zeroing on Marshal and Cleanup, effective-value resolution for overlapping fields).

Smaug123:
Superseded by #819, which is the same feature built the other way round.

This PR drove the stub by re-executing a `calli`, recovering progress by counting values on
CoreLib's evaluation stack, because a stub had no `MethodInfo` and so could not have a frame. The
design note above records why that was provisional: the idiom is only sound on an *owned, empty*
stack, and `DateMarshaler.ConvertToNative` is the only field marshaller in CoreLib returning a
non-void value, so a void-returning conversion could not have been counted at all.

Giving the stub a synthesised `MethodInfo` — and therefore a real frame — turned out to *delete*
more of this PR than it kept. #819 has no `NativeIntSource.StructMarshalStub` case and none of its
30 exhaustive-match sites, no `calli` fork, no call-site arity assertion, no program-counter
assertion, and no manual `SuspendedForClassInit` stack restoration. `calli` is untouched there.
So the driver here was not "good code with a provisional resumption mechanism"; it was scaffolding
that existed solely to work around not having a frame, plus permanent API surface on
`NativeIntSource` that `main` would have carried and then had to remove.

What survives into #819 unchanged: the marshalling semantics (OADate via the guest's own helper,
whole-image zeroing on Marshal and Cleanup, the structural source read, effective-value resolution
for overlapping fields, `[MarshalAs]` refusal) and all four differential test sources. The four
review passes on this PR — two workflow, two Codex — were therefore spent almost entirely on
things that carried over; the findings they produced are live in #819.

The layout-walk refactor from this PR landed separately as #811.

Branch retained for reference.

PR #806, opened 2026-08-08
Separate a method's universal facts from its metadata-backed ones

Smaug123:
**Stacked on #803** — review that first; this PR's own diff is the last commit,
`Separate a method's universal facts from its metadata-backed ones`.

Groundwork for letting the runtime synthesise methods. It is deliberately inert: nothing
constructs a `Synthesised` yet, so the change is behaviour-preserving by construction, and the
first consumer (porting the struct-marshal stub onto a real frame) follows separately.

## Why

#803's struct-marshal stub is driven by re-executing a `calli`, recovering "how many conversions
have completed" by counting values on CoreLib's evaluation stack. That works, and is tested, but
it is a local optimum: the idiom is borrowed from `NativeRuntimeTypeQCall`, where it is sound
because a *native frame* has its own empty evaluation stack. Transplanted onto a frame owned by
`Marshal.StructureToPtr`, it is unambiguous only by inspection. It also does not generalise —
`DateMarshaler.ConvertToNative` is the only field marshaller in CoreLib returning a non-void
value, and a conversion that returns nothing cannot be counted by a mechanism that counts stack
values. See the design note on #803 for the full argument.

The fix is to give such a stub a real frame, which needs a `MethodInfo` that no MethodDef row
backs. Three other parked features want the same thing: the JIT-helper allocator
`MarshalPtrToStructure.cs` is parked on ("CoreCLR's allocator is a JIT helper ... with no managed
MethodInfo to put in `NativeIntSource.FunctionPointer`"), delegate `BeginInvoke`/`EndInvoke`
currently sitting in `RuntimeBehaviour.Unrecognised`, and the multi-dim array helpers
`MethodBody`'s doc comment anticipates. So the capability is worth building on its own terms.

## What

`MethodInfo` becomes a two-case DU over a shared `MethodCore`:

- The **six facts every method has** — declaring type, name, body, generics, signature,
  staticness — are projected as members. The several hundred sites that only read those are
  untouched, which is what kept this to ~60 real decision points rather than a 500-site rewrite.
- The **seven that exist only because a MethodDef row does** — token, Param rows, raw signature,
  custom attributes, method and impl attributes, native import — are deliberately *not*
  projected. Reaching them requires matching.

That asymmetry is the whole point. The alternative considered was widening the identity field
alone and letting synthesised methods carry an empty Param collection and zeroed attribute flags;
that is the same class of lie as the fabricated `MethodDefinitionHandle` this design exists to
avoid, just spread thinner across more fields.

**Where the line falls on projection.** A *predicate with a genuine answer* gets a member:
`IsVirtual`, `IsNewSlot`, `IsFinal`, `IsPublic`, `TryNativeImport` are all false/None for a
synthesised method because it occupies no vtable slot, overrides nothing, and is reachable only
from the interpreter that synthesised it. Those are true answers, not fabricated ones. The *raw
attribute flags* stay behind a match, because handing the guest a zero would be inventing
metadata — so the two reflection paths that report them (`RuntimeMethodHandle.GetAttributes`, the
custom-attribute constructor lookup) now fail loudly instead, as does minting a
`RuntimeMethodHandle` for a synthesised method.

Stack-trace rendering is the one place that degrades rather than refuses: CoreCLR's own IL stubs
do appear in stack traces, so a synthesised frame renders without parameter detail rather than
killing the run.

## Incidental

The split surfaced a dozen call sites using `Parameters.Length`/`IsEmpty` as an arity check —
which that field's own doc comment warns against at length, because the Param table under-counts
(F# emits abstract members with zero Param rows regardless of declared arity). Those now use
`MethodInfo.arity`.

Two tests genuinely mean the Param table rather than the arity — *"AbstractDispatch's abstract
Combine really does have zero Param rows"* and its byref sibling — and they caught the sweep that
had converted them into the thing they are contrasted against. They keep reading the raw field,
with a comment saying why. Worth noting the compiler could not have caught that one; only running
the suite could.

### DIFF
--- a/WoofWare.PawPrint.App/DebuggerServer.fs
+++ b/WoofWare.PawPrint.App/DebuggerServer.fs
@@ -722,10 +722,15 @@ module DebuggerServer =
             // The executing method has been concretised, so its own generic parameters no longer
             // carry their declared names; recover them from the metadata flavour of the same
             // method. If it isn't in the index, the indices render positionally.
+            // A synthesised frame has no metadata flavour to recover names from, so it renders
+            // positionally too.
             let scope =
-                match assembly.Methods.TryGetValue frame.ExecutingMethod.Handle with
-                | true, method -> GenericScope.ofMethod method
-                | false, _ -> GenericScope.unknown
+                match frame.ExecutingMethod.TryMetadata with
+                | None -> GenericScope.unknown
+                | Some facts ->
+                    match assembly.Methods.TryGetValue facts.Handle with
+                    | true, method -> GenericScope.ofMethod method
+                    | false, _ -> GenericScope.unknown
 
             writer.WriteStartObject ()
             writer.WriteNumber ("thread", threadIdValue threadId)
--- a/WoofWare.PawPrint.Domain/Assembly.fs
+++ b/WoofWare.PawPrint.Domain/Assembly.fs
@@ -676,7 +676,17 @@ module Assembly =
         // TODO: this probably misses any methods out which aren't associated with a type definition?
         let methods =
             typeDefs
-            |> Seq.collect (fun (KeyValue (_, ty)) -> ty.Methods |> List.map (fun mi -> KeyValuePair (mi.Handle, mi)))
+            |> Seq.collect (fun (KeyValue (_, ty)) ->
+                ty.Methods
+                // Keyed by MethodDef token, so only metadata-backed methods belong in it. Nothing
+                // synthesised exists at read time — this dictionary is built straight out of the
+                // PE image — but the match keeps the key honest rather than inventing a token.
+                |> List.choose (fun mi ->
+                    match mi with
+                    | MethodInfo.Metadata (_, facts) -> Some (KeyValuePair (facts.Handle, mi))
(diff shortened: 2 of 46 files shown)


Smaug123:
Superseded by #819, which contains this change as its first commit.

Nothing here was wrong — the identity split is carried over essentially verbatim, rebased onto
`main` and with two extra call sites that arrived in the meantime (`AppContextSeed`, and one more
in `TestMethodHandleRegistry`).

What changed is the sequencing. This PR was based on #803 rather than on `main`, purely because it
was written after it; and it landed the `SynthesisedMethod` DU with no producer, which made a
one-case DU look like speculative generality. #819 lands the split together with its first
consumer — the struct-marshal stub — which is the shape that justifies it, and does so against
`main` so the identity work is not gated behind a marshalling feature.

The three other parked features that want synthesised methods are unaffected either way: the
JIT-helper allocator `MarshalPtrToStructure.cs` is parked on, delegate `BeginInvoke`/`EndInvoke`,
and the multi-dim array helpers.

Branch retained for reference.

PR #924, opened 2026-08-10
Snapshot a guest at its first contended scheduling decision

Smaug123:
Follows #918 and #919, both merged. This is the payoff stage of `docs/plans/2026-08-10-fork-point-snapshots.md`.

Sweeping PCT seeds over one guest re-executes an identical single-threaded startup for every seed — measured at 74–94% of a run's instructions and ~90% of its wall clock. Now that the scheduler only mutates its policy state at a contended decision (#918), that prefix is provably the same execution under every seed, so it can be computed once and each seed resumed from it.

```fsharp
val runToFirstFork : ILoggerFactory -> string option -> Stream -> GuestConfig -> PrefixOutcome
val runToNextFork  : ILoggerFactory -> ILogger -> PreparedProgram -> PrefixOutcome
val resumeFork     : ILoggerFactory -> uint64 option -> ForkSnapshot -> PreparedProgram
```

`resumeFork` hands back an ordinary `PreparedProgram`, so callers drive it with the existing `stepPrepared` / `pumpPrepared`. Wiring the sweeps up to it is stage 4 and is not in this PR.

## Design notes

* **The snapshot is an inter-tick value; the probe is not.** Contention can be created *inside* a tick — a deadline firing, a spurious wake, the signal dispatcher waking — so `stepPrepared` is split into `advanceToDecision` and `stepDecided`, and the probe runs after the preamble. But the state *kept* is from before it: a mid-tick value would be a new kind of resumable thing that the ordinary driver would double-run the preamble on, advancing `StepCounter` twice and shifting the spurious-wakeup schedule. Resuming replays that tick's preamble instead, which is policy-independent and so reproduces it exactly.
* **`runToNextFork` is the general primitive**, so the schedule-space tree search this is eventually for needs no rework. Only what *resume* means differs between a first fork and a mid-run one, and `IlMachineState.withPctSeed` now spells that difference out: re-seeding mid-run is "re-randomise the future", not schedule replay.
* **`Scheduler.isContended` became `tryContenders`**, returning the witness. The fork runner needs to name the contenders, and deriving that list separately would be a second definition of the predicate, free to disagree with the first.
* **A fork inside a `.cctor` is detected and refused**, not snapshotted: resuming one needs a two-shape return value that no caller in this repo would use. The extension point is one DU arm.
* `resumeFork` rebinds the state's logger, which would otherwise remain the prefix's and silently drop every per-seed log property.


WoofWare.PawPrint/Scheduler.fs
```diff
@@ -119,20 +119,27 @@ module Scheduler =
...
+    /// policy may consume randomness at this tick, and iff this tick is a fork point in the
+    /// schedule space.
     ///
-    /// Exposed for harnesses that want to run a guest up to its first fork point and then fan
-    /// out over seeds from there. Must be evaluated at the same moment the scheduler would make
-    /// its choice — i.e. after the driver's per-tick preamble (spurious wakeups, deadline
-    /// firing, signal-handler spawn, deadline jump), any of which can create contention within
-    /// the tick that a probe on the inter-tick state would miss.
-    let isContended (state : IlMachineState) : bool =
+    /// Returns the witness rather than a bare `bool` so that a caller which needs to *name* the
```
Smaug123: ugh slop comment

